PR #37, opened 2024-05-15
optimize css output by inlining dynamic values

radnan:
taking inspiration from https://github.com/wayfair/rainbow-sprinkles/ to inline dynamic values rather than generating the full list of pre-defined combinations

this will greatly reduce the final css bundle size since we can just inline the css vars needed by each css rule

#### PR
we've now enabled `conditions` and `modifiers` for all sprinkle props - and we actually **reduced** the css file size by `39KB`:

image

#### Main
if we tried to do this in `main` branch with previous implementation the css size would **increase** by `158KB`:

image

#### PR
and if we added one more modifier here (for example `&[data-disabled]`) the size would only just increase by another `11KB` in this PR:

image

#### Main
whereas in `main` the size would increase by `74KB`:

image


packages/react/src/sprinkles/sprinkles.css.ts
```diff
@@ -55,235 +55,219 @@ const modifiers = {
...
-    conditions,
-    properties: {
-      alignItems: ["center", "end", "normal", "start", "stretch"] as const,
-      colSpan: {
-        "1": { gridColumn: "span 1 / span 1" },
-        "2": { gridColumn: "span 2 / span 2" },
+export const props = defineProperties({
+  "@layer": layers.axiom,
+  conditions,
+  modifiers,
```
radnan, line 61: The type declaration output has ballooned in size because of combining everything - will need to investigate how to reduce that

MunifTanjim (approved):
I don't understand most of it 😂 LGTM 👍🏼

PR #40, opened 2024-05-20
Button Component

JafarSaifullah:
## Description
Button component is developed in this PR. The following items are generated for the Button Component:
- Recipes
- Storybook
- Package (Component) 

## QA Steps
- [ ] Check Button Stories to see all the possible buttons.

#### Before Merging

- [ ] 

#### After Deployed

- [ ]

### DIFF
--- a/packages/react/src/button/Button.tsx
+++ b/packages/react/src/button/Button.tsx
@@ -15,55 +15,89 @@ import { Flex } from "../flex";
 import { Text } from "../text";
 import { type Recipe, recipe } from "./Button.recipe";
 
+const presets = {
+  basic: { colorScheme: "secondary", variant: "solid" },
+  "basic-link": { colorScheme: "secondary", variant: "link" },
+  danger: { colorScheme: "danger", variant: "solid" },
+  "danger-outline": { colorScheme: "danger", variant: "outline" },
+  plain: { colorScheme: "secondary", variant: "ghost" },
+  primary: { colorScheme: "primary", variant: "solid" },
+} satisfies Record<string, Recipe>;
+
 type ButtonProps = ExtendProps<
   ComponentPropsWithRef<"button">,
   ComponentPropsWithRef<typeof Flex>,
   {
     asChild?: boolean;
     children?: ReactNode;
+    isLoading?: boolean;
     leftSection?: ReactNode;
+    preset?: keyof typeof presets;
     rightSection?: ReactNode;
   } & Recipe
 >;
 
 export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
-  ({ asChild, children, leftSection, rightSection, ...props }, ref) => {
+  (
+    {
+      asChild,
+      children,
+      colorScheme,
+      disabled,
+      isLoading,
+      leftSection,
+      onClick,
+      preset = "basic",
+      rightSection,
+      variant,
+      ...props
+    },
+    ref,
+  ) => {
     const Comp = asChild ? Slot : "button";
     children =
       asChild && isValidElement(children) ? (
         cloneElement(
           children,
           undefined,
-          <Text fontSize="inherit" fontWeight="600">
-            {children.props.children}
-          </Text>,
+          <Text fontSize="inherit">{children.props.children}</Text>,
         )
       ) : (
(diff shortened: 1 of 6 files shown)


packages/react/src/button/Button.tsx
```diff
@@ -15,38 +15,70 @@ import { Stack } from "../stack";
 import { Text } from "../text";
 import { type Recipe, recipe } from "./Button.recipe";
 
+const presets = {
+  basic: { colorScheme: "secondary", variant: "solid" },
+  basicLink: { colorScheme: "secondary", variant: "link" },
```
radnan, line 20: let's use dash case for prop values
```suggestion
  "basic-link": { colorScheme: "secondary", variant: "link" },
```

packages/react/src/button/Button.tsx
```diff
@@ -15,38 +15,70 @@ import { Stack } from "../stack";
...
+  danger: { colorScheme: "danger", variant: "solid" },
+  dangerOutline: { colorScheme: "danger", variant: "outline" },
```
radnan, line 22: let's do dash case here as well
```suggestion
  "danger-outline": { colorScheme: "danger", variant: "outline" },
```

packages/react/src/sprinkles/sprinkles.css.ts
```diff
@@ -35,6 +35,7 @@ const conditions = {
 } as const;
 
 const states = {
+  active: "&:active",
```
radnan, line 38: let's drop the commit with these changes to `sprinkles.css.ts` since they exist in `main` now

packages/react/src/css-idents.ts
```diff
@@ -341,4 +341,7 @@ export default {
   c7: "data-disabled",
   c8: "hover-enabled",
   c9: "aria-error",
+  ca: "data-disabled",
```
radnan, line 344: let's drop the commit with these changes to `css-idents.ts` since they exist in `main` now

packages/react/src/button/Button.tsx
```diff
@@ -58,12 +90,24 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+          colorScheme: finalColorScheme,
+          variant: finalVariant,
+        })}
+        data-disabled={props.disabled}
       >
         <Comp ref={composedRef}>
-          {leftSection}
-          <Slottable>{children}</Slottable>
-          {rightSection}
+          {isLoading ? (
```
radnan, line 101: we can't do this otherwise `Slot` will break

let's address this in a separate PR completely and change this back to how it was before

packages/react/src/button/Button.tsx
```diff
@@ -15,38 +15,70 @@ import { Flex } from "../flex";
...
-          {children}
-        </Text>
+        <Text fontSize="inherit">{children}</Text>
       );
 
     const innerRef = useRef<HTMLButtonElement>(null);
     const composedRef = useComposedRefs(ref, innerRef);
 
+    if (isLoading) {
+      props.disabled = true;
```
radnan, line 76: this is not a good pattern - technically `props` is meant to be a read only variable. plus we don't want to forward `disabled` to the element either.

so instead we should extract these 2 props out in the function signature and use them accordingly:

```
const Button = ({ disabled, isLoading, ...props }) => (
  <button data-disabled={disabled || isLoading} {...props} />
)
```

packages/react/src/button/Button.tsx
```diff
@@ -15,38 +15,70 @@ import { Flex } from "../flex";
...
+  { colorScheme: Recipe["colorScheme"]; variant: Recipe["variant"] }
+>;
+
 type ButtonProps = ExtendProps<
   ComponentPropsWithRef<"button">,
   ComponentPropsWithRef<typeof Flex>,
   {
     asChild?: boolean;
     children?: ReactNode;
+    colorScheme?: Recipe["colorScheme"];
```
radnan, line 36: we don't need to duplicate these since we already intersected the `Recipe` type with `& Recipe`:

Screenshot 2024-05-29 at 2 20 00 PM
```suggestion
```

radnan (approved):

PR #41, opened 2024-05-21
Text Input design

toha993:
Added text input design and stories


packages/react/src/input/Input.tsx
```diff
@@ -0,0 +1,46 @@
...
+        data-disabled={isDisabled}
+        display="flex"
+        flexDirection="column"
+        fontFamily="sans"
+        fontSize="md"
+        maxW="sm"
+        {...recipe({ size, variant })}
+        {...props}
+      >
+        <input placeholder={placeholder} ref={ref} type={type} />
```
radnan, line 40: let's spread the `...props` onto the `<input />` component instead - so we will pick certain props like `className` to be on the outer wrapper/box - but everything else will get forwarded to inner input element

packages/react/src/input/Input.tsx
```diff
@@ -0,0 +1,46 @@
...
+      ...props
+    },
+    ref,
+  ) => {
+    return (
+      <Box
+        aria-invalid={isInvalid}
+        asChild
+        data-disabled={isDisabled}
+        display="flex"
```
radnan, line 32: let's add `alignItems: 'center'` here in addition

packages/react/src/input/Input.tsx
```diff
@@ -0,0 +1,50 @@
...
+      isInvalid,
+      placeholder,
+      size = "md",
+      type,
+      variant = "default",
+      ...props
+    },
+    ref,
+  ) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
```
MunifTanjim, line 29: Is this needed?

packages/react/src/input/Input.css.ts
```diff
@@ -0,0 +1,97 @@
...
+          [layers.axiom]: {
+            textAlign: "end",
+          },
+        },
+      },
+    },
+  },
+});
+
+export type InpuVariants = RecipeVariants<typeof input>;
```
MunifTanjim, line 97:
```suggestion
export type InputVariants = RecipeVariants<typeof input>;
```

packages/react/src/input/Input.tsx
toha993: Yes Vai We will give the sprinkle props to the Box and pass rest of the props to the input elements like `onChange` etc.

packages/react/src/input/Input.css.ts
toha993: Got it vai.

packages/react/src/input/Input.tsx
MunifTanjim: Doesn't it happen automatically?

```diff
diff --git a/packages/react/src/input/Input.tsx b/packages/react/src/input/Input.tsx
index d9dd2dd..0754f0a 100644
--- a/packages/react/src/input/Input.tsx
+++ b/packages/react/src/input/Input.tsx
@@ -2,7 +2,6 @@ import clsx from "clsx";
 import { type ComponentPropsWithRef, forwardRef } from "react";
 
 import { Box } from "../box";
-import { extractSprinkles } from "../sprinkles";
 import { type ExtendProps } from "../utils";
 import { type InpuVariants, input } from "./Input.css";
 
@@ -26,22 +25,15 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
     },
     ref,
   ) => {
-    const { restProps, sprinkleProps } = extractSprinkles(props);
     return (
       <Box
         aria-disabled={isDisabled}
         aria-invalid={isInvalid}
         asChild
-        className={clsx(
-          input({
-            size,
-            variant,
-          }),
-          className,
-        )}
-        {...sprinkleProps}
+        className={clsx(input({ size, variant }), className)}
+        {...props}
       >
-        <input placeholder={placeholder} ref={ref} type={type} {...restProps} />
+        <input placeholder={placeholder} ref={ref} type={type} />
       </Box>
     );
   },
```

this is working for me.

packages/react/src/input/Input.css.ts
```diff
@@ -0,0 +1,97 @@
...
+          },
+        },
+      },
+      number: {
+        "@layer": {
+          [layers.axiom]: {
+            textAlign: "end",
+          },
+        },
+      },
```
MunifTanjim, line 92: @radnan vai, I'm very hesitant on adding this.

Do we really need to add this right now?

toha993: I believe it's best to give the power to end user where they want to align rather than forcing it vai.  I also don't like this as we need to expose variant for this.

MunifTanjim: It's not even about giving power. To me it just seems like adding bloat to the library... 🤷‍♂️ And the thing about adding something is, once it's being used somewhere you can't remove it easily anymore.

packages/react/src/input/Input.tsx
radnan: It happens automatically now but once we add the wrapper to accommodate the left/right sections it won’t work anymore since we’ll add an extra wrapper

packages/react/src/input/Input.tsx
radnan: See https://github.com/optimizely-axiom/optiaxiom/pull/41#discussion_r1629601071

radnan (approved):

packages/react/src/input/Input.tsx
toha993: @JafarSaifullah

PR #47, opened 2024-05-27
Input with icon design

toha993:


packages/react/src/input/Input.tsx
```diff
@@ -27,11 +27,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
         w="256"
         {...ParentBoxrecipe(props)}
       >
-        {leftSection && (
-          <Box >
-            {leftSection}
-          </Box>
-        )}
```
MunifTanjim: I'm not sure what editor setup you're using... but these should never be separate commit. You should always commit linted and formatted codes.

toha993: Got it vai.

packages/react/src/input/Input.tsx
```diff
@@ -9,16 +9,25 @@ import * as styles from "./Input.css";
...
+  } & styles.WrapperVariants
 >;
 
 export const Input = forwardRef<HTMLInputElement, InputProps>(
   (
     {
       className,
+      defaultValue,
+      id,
       isDisabled,
```
radnan, line 26: let's rename `isDisabled` to just `disabled` - since that's the actual attribute name. that way we're following the general convention, not making it a surprise to users, and making it consistent with other components like `Button`.

packages/react/src/input/Input.tsx
```diff
@@ -9,16 +9,25 @@ import * as styles from "./Input.css";
...
       isInvalid,
```
radnan, line 27: let's rename this from `isInvalid` to `error`. there is not `error` html attribute - but this is just to remain consistent with other attributes like `disabled`, `readonly`, etc

packages/react/src/input/Input.tsx
```diff
@@ -43,7 +51,34 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
...
+          className={clsx(
+            styles.input({
+              variant,
+            }),
+          )}
+        >
+          <input
+            defaultValue={defaultValue}
+            id={id}
+            placeholder={placeholder}
```
radnan, line 70: are we actually doing anything with the `placeholder`? if not let's just forward it via `restProps` instead of picking it out and passing it explicitly

packages/react/src/input/Input.tsx
```diff
@@ -43,7 +51,34 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
...
+            {leftSection}
+          </Box>
+        )}
+        <Box
+          asChild
+          className={clsx(
+            styles.input({
+              variant,
+            }),
+          )}
```
radnan, line 65: let's shorten this to a single line
```suggestion
          className={styles.input({ variant })}
```

and remove `clsx` since we're not merging any classes here

packages/react/src/input/Input.tsx
toha993: Before restProps were introduced, I needed to pass the prop explicitly for the placeholder. With restProps, we wouldn't need it anymore vai.

packages/react/src/input/Input.tsx
toha993: Sure vai

packages/react/src/input/Input.tsx
toha993: Okay vai

packages/react/src/input/Input.tsx
toha993: Okay vai

radnan (approved):

PR #55, opened 2024-05-29
`Field` component

toha993:


apps/storybook/src/forms/FormGroup.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+      </FormField>
+
+      <FormField disabled label="Form label">
+        <Input placeholder="Disabled" />
+      </FormField>
+
+      <FormField error="Form note" label="Form label">
+        <Input placeholder="With error" />
+      </FormField>
+    </Flex>
```
radnan, line 38: let's split these into separate stories rather than rolling them all into one

otherwise it's hard to figure out what's the intent for each at a glance

packages/react/src/form-field/FormField.css.ts
```diff
@@ -0,0 +1,33 @@
...
+export const labelRequired = style({
+  color: theme.colors["border.error"],
+});
+
+export const error = style({
+  color: theme.colors["border.error"],
+  fontSize: "12px",
+  lineHeight: "20px",
+  marginTop: "2px",
+});
```
radnan, line 33: let's not hard code font sizes

for this whole file - since we're using very simple styles we can just replace them all with sprinkles

```suggestion
export const error = sprinkles({
  color: "border.error",
  fontSize: "md"
  mt: "2",
});
```

in fact if we're just using sprinkles and not doing any complicated styles, let's inline the styles into the tsx file rather than creating a separate `.css.ts` file

packages/react/src/form-field/FormField.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const FormField = forwardRef<HTMLDivElement, FormGroupProps>(
+  (
+    {
+      children,
+      className,
+      description,
+      disabled,
+      error,
+      id: _id,
```
radnan, line 37: let's name this `idProp`
```suggestion
      id: idProp,
```

so it's clear that this is a prop value that we've renamed for a particular reason

packages/react/src/form-field/FormField.tsx
```diff
@@ -0,0 +1,74 @@
...
+    },
+    ref,
+  ) => {
+    const id = useId(_id);
+    return (
+      <Box className={clsx(styles.formField, className)} ref={ref} {...props}>
+        {label && (
+          <RadixLabel.Root className={styles.label}>
+            {label}{" "}
+            {required && <span className={styles.labelRequired}>*</span>}
```
radnan, line 50: let's append the space to the asterisk
```suggestion
            {label}
            {required && <span className={styles.labelRequired}> *</span>}
```

radnan, line 50: let's use [`<VisuallyHidden />`](https://www.radix-ui.com/primitives/docs/utilities/visually-hidden) for the asterisk

instead of just a simple span

packages/react/src/form-field/FormField.tsx
```diff
@@ -0,0 +1,74 @@
...
+      required,
+      ...props
+    },
+    ref,
+  ) => {
+    const id = useId(_id);
+    return (
+      <Box className={clsx(styles.formField, className)} ref={ref} {...props}>
+        {label && (
+          <RadixLabel.Root className={styles.label}>
```
radnan, line 48: we need an `htmlFor` here

packages/react/src/form-field/FormField.tsx
```diff
@@ -0,0 +1,74 @@
...
+    description?: string;
+    disabled?: boolean;
+    error?: string;
+    id?: string;
+    label?: string;
+    required?: boolean;
+  }
+>;
+
+export const FormField = forwardRef<HTMLDivElement, FormGroupProps>(
```
radnan, line 29: let's simplify this name to just `Field` - no need for the `Form` prefix

packages/react/src/form-field/FormField.tsx
radnan: brain fart - i meant hide the asterisk from screen readers

apps/storybook/src/forms/Field.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+export default meta;
+
+type Story = StoryObj<typeof Field>;
+
+export const InputWithIcon: Story = {
+  render: () => (
+    <Field label="Form label" required>
+      <Input leftSection={<IconCalendar size="16" />} placeholder="With Icon" />
+    </Field>
+  ),
```
radnan, line 22: let's use the `args` convention instead of `render` - that way the controls in storybook will be integrated into the story

```suggestion
  args: {
    children: <Input leftSection={<IconCalendar size="16" />} placeholder="With Icon" />,
    label: 'Form label',
    required,
  },
```

https://github.com/optimizely-axiom/optiaxiom/assets/758244/92f235b4-2e32-48ab-8522-4af7175cc124

packages/react/src/field/Field.tsx
```diff
@@ -0,0 +1,85 @@
...
+          id,
+          required,
+        })}
+        {description && (
+          <Text as="p" color="fg.default" fontSize="sm" mt="2">
+            {description}
+          </Text>
+        )}
+        {error && (
+          <Text as="p" color="border.error" fontSize="sm" mt="2">
```
radnan, line 76: shouldn't this be `fg.error` ?
```suggestion
          <Text as="p" color="fg.error" fontSize="sm" mt="2">
```

packages/react/src/field/Field.tsx
```diff
@@ -0,0 +1,85 @@
...
+          </RadixLabel.Root>
+        )}
+        {cloneElement(children, {
+          disabled,
+          error: !!error,
+          id,
+          required,
+        })}
+        {description && (
+          <Text as="p" color="fg.default" fontSize="sm" mt="2">
```
radnan, line 71: let's use a flexbox here and use `gap` - rather than controlling the margin of each individual element

packages/react/src/field/Field.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ...props
+    },
+    ref,
+  ) => {
+    const id = useId(idProp);
+    return (
+      <Box display="flex" flexDirection="column" maxW="sm" ref={ref} {...props}>
+        {label && (
+          <RadixLabel.Root
+            className={sprinkles({
```
radnan, line 47: let's use `<Text asChild>` here instead of manually using `className + sprinkles`

radnan: i guess as a general rule `sprinkles()` should only really be used in `.css.ts` files - and in tsx we should use the native props

radnan (approved):

PR #56, opened 2024-05-29
Add `Textarea` Component

JafarSaifullah:
## Description
`Textarea` component is developed in this PR. The following items are generated for the Button Component:
- Recipes
- Storybook
- Package (Component) 

## QA Steps
- [ ] Check TextArea Stories.


packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type TextareaProps = ExtendProps<
+  ComponentPropsWithRef<"textarea">,
+  ComponentPropsWithRef<typeof Box>,
+  {
+    autosize?: boolean;
+    bottomSection?: ReactNode;
+    defaultValue?: string;
+    isDisabled?: boolean;
+    isInvalid?: boolean;
```
radnan, line 18: let's rename `isDisabled` to `disabled`

and rename `isInvalid` to `error`

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,77 @@
...
+    styles.TextAreaRecipeVariants
+>;
+
+export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
+  (
+    {
+      autosize,
+      bottomSection,
+      className,
+      defaultValue,
```
radnan, line 31: no need to pick out `defaultValue` since we're just forwarding it to the textarea

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,77 @@
...
+type TextareaProps = ExtendProps<
+  ComponentPropsWithRef<"textarea">,
+  ComponentPropsWithRef<typeof Box>,
+  {
+    autosize?: boolean;
+    bottomSection?: ReactNode;
+    defaultValue?: string;
+    isDisabled?: boolean;
+    isInvalid?: boolean;
+    isResizeable?: boolean;
```
radnan, line 19: having both `autosize` and `isResizeable` now feels a bit confusing

can we combine these two into a single prop called `resize: 'auto' | 'vertical' | 'none'`

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,53 @@
+import { recipe } from "../vanilla-extract";
+
+export const textAreaBoxRecipe = recipe({
```
radnan, line 3: let's rename this to 
```suggestion
export const textarea = recipe({
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,53 @@
...
+    selectors: {
+      "&:focus-visible": {
+        outline: "0",
+      },
+    },
+  },
+  variants: {},
+});
+
+export const parentBoxRecipe = recipe({
```
radnan, line 18: let's rename this to
```suggestion
export const wrapper = recipe({
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,53 @@
+import { recipe } from "../vanilla-extract";
+
+export const textAreaBoxRecipe = recipe({
+  base: {
+    background: "transparent",
+    flexGrow: "1",
+    fontFamily: "sans",
```
radnan, line 7: we don't need this

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,53 @@
...
+      },
+      "&:data-disabled": {
+        backgroundColor: "bg.disabled",
+        borderColor: "border.secondary",
+        color: "fg.quaternary",
+        pointerEvents: "none",
+      },
+      "&:focus-within": {
+        outline: "2",
+        outlineColor: "brand.200",
```
radnan, line 41: we can't refer to design tokens directly in recipe - it has to be through `theme` or via `sprinkles()`

```suggestion
        outlineColor: theme.colors["brand.200"],
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,53 @@
...
+      "&:hover": {
+        borderColor: "border.brand",
+      },
+    },
+  },
+  variants: {},
+});
+
+export type ParentRecipeVariants = Parameters<typeof parentBoxRecipe>;
+export type TextAreaRecipeVariants = Parameters<typeof textAreaBoxRecipe>;
```
radnan, line 53:
```suggestion
export type WrapperVariants = Parameters<typeof parentBoxRecipe>;
export type TextareaVariants = Parameters<typeof textAreaBoxRecipe>;
```

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: addressed

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: addressed bhai

packages/react/src/textarea/Textarea.css.ts
JafarSaifullah: addressed

radnan, line 53: looks like this is the problem area - these need to be `RecipeVariants` and not just `Parameters` (cause then it becomes an array of the arguments)

mostafa-K-raihan:
will this component support callbacks, like onChange?
Also, we might need to call some APIs/save some states when we press `Enter`/go off click, will this component support passing `onKeyDown`/off click?

JafarSaifullah:
We are extending the default 'textarea' component and also the `Box` component. So this will support the required callbacks

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: Bhai, I think we should keep two props - `AutoResize` and `resize`. 

1. `AutoResize` is a boolean value which will provide the functionality of auto resizing the Component.
2. `resize` is the prop that lets the user expand functionality using the bottom right 'expand' icon.  image
 
 
 I can use a single prop to render the same functionality but I think having two separate props make sense in this issue bhai.
 
We can make the following change to resize prop - `resize: 'vertical' | 'none'`.

packages/react/src/textarea/Textarea.css.ts
JafarSaifullah: keeping flexGrow as the middle part of textarea should expand if needed, bhai

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,80 @@
...
+export const wrapper = recipe({
+  base: [
+    {
+      border: "1",
+      fontSize: "md",
+      overflow: "auto",
+      rounded: "sm",
+    },
+    style({
+      border: "1",
```
toha993, line 14: Duplicate styling

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,80 @@
...
+export const textarea = recipe({
+  base: [
+    {
+      w: "full",
+    },
+    style({
+      alignItems: "flex-start",
+      borderColor: theme.colors["border.default"],
+      color: theme.colors["fg.default"],
+      flexGrow: "1",
```
toha993, line 65: We have `flex` for this property also.

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,79 @@
...
+  (
+    {
+      autoResize = false,
+      bottomSection,
+      className,
+      disabled,
+      error,
+      placeholder,
+      readOnly,
+      resize = "ve",
```
toha993, line 35: I think it misspelled with vertical

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,79 @@
...
+        {...styles.wrapper()}
+        {...sprinkleProps}
+      >
+        {topSection && <Box>{topSection}</Box>}
+        <Box
+          asChild
+          {...styles.textarea({}, className)}
+          style={{
+            color: disabled ? "fg.disabled" : "fg.default",
+          }}
```
toha993, line 63:
```suggestion
 color={disabled ? "fg.disabled" : "fg.default"}
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,80 @@
...
+          outlineColor: theme.colors["red.200"],
+          outlineOffset: "1px",
+          outlineStyle: "solid",
+          outlineWidth: "2px",
+        },
+        '&:focus-within:not([data-invalid="true"])': {
+          outlineColor: theme.colors["brand.200"],
+          outlineOffset: "1px",
+          outlineStyle: "solid",
+          outlineWidth: "2px",
```
toha993, line 38:
```suggestion
          outline: `2px solid ${theme.colors["brand.200"]}`,
          outlineOffset: "1px",
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,80 @@
...
+        "&:focus-within": {
+          outline: "2",
+          outlineColor: theme.colors["brand.200"],
+          outlineOffset: "1",
+        },
+        '&:focus-within:is([data-invalid="true"])': {
+          outlineColor: theme.colors["red.200"],
+          outlineOffset: "1px",
+          outlineStyle: "solid",
+          outlineWidth: "2px",
```
toha993, line 32:
```suggestion
        '&:focus-within:is([data-invalid="true"])': {
          outline: `2px solid ${theme.colors["red.200"]}`,
          outlineOffset: "1px",
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,80 @@
...
+      border: "1",
+      fontSize: "md",
+      overflow: "auto",
+      rounded: "sm",
+    },
+    style({
+      border: "1",
+      borderColor: theme.colors["border.default"],
+      gap: theme.spacing["xs"],
+      letterSpacing: "-.14px",
```
toha993, line 17: Do we need this?

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,80 @@
+import { theme } from "../styles";
+import { style } from "../vanilla-extract";
+import { recipe } from "../vanilla-extract";
```
toha993, line 3: We can import it in the same line

```suggestion
import { recipe, style } from "../vanilla-extract";
```

apps/storybook/src/forms/Textarea.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+export const Custom: Story = {
+  args: {
+    autoResize: true,
+    bottomSection: (
+      <Flex
+        style={{
+          alignItems: "stretch",
+          justifyContent: "center",
+        }}
```
toha993, line 69:
```suggestion
          alignItems: "stretch",
          justifyContent: "center",
```

packages/react/src/textarea/Textarea.css.ts
JafarSaifullah: I coded it based on the figman file
alt text

toha993: Not sure if we need it or not. 
C.C @radnan vai.

packages/react/src/textarea/Textarea.css.ts
JafarSaifullah: Addressed

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: my bad. I missed it

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,72 @@
...
+      gap: "xs",
+      overflow: "auto",
+      p: "xs",
+      rounded: "sm",
+      w: "340",
+    },
+    style({
+      letterSpacing: "-0.14px",
+      selectors: {
+        "&:data-invalid": {
```
toha993, line 18:
```suggestion
        '&[data-invalid="true"]': {
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,72 @@
...
+          borderColor: "border.error",
```
toha993, line 19: Its already in 41st line. No need of this anymore.

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    return (
+      <Flex
+        aria-disabled={disabled}
+        aria-invalid={error}
+        data-disabled={disabled}
+        data-invalid={error}
+        style={{
+          resize: resize === "none" ? "none" : "vertical",
+        }}
```
toha993, line 53: Can we add style with 
```
export const resizeNone = recipe({
  base: [
    style({
      resize: "none",
    }),
  ],
});

export const resizeVertical = recipe({
  base: [
    style({
      resize: "vertical",
    }),
  ],
});

```

and use clsx to merge the style instead of using style in tsx?

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type TextareaProps = ExtendProps<
+  ComponentPropsWithRef<"textarea">,
+  ComponentPropsWithRef<typeof Box>,
+  {
+    autoResize?: boolean;
+    bottomSection?: ReactNode;
+    disabled?: boolean;
+    error?: boolean;
+    readonly?: boolean;
```
toha993, line 19: Text area already has read only props. We don't need it explicitly.

JafarSaifullah: addressed

packages/react/src/textarea/Textarea.css.ts
JafarSaifullah: missed it

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: I guess, this is simpler in this manner.

radnan: we should never use inline `styles` unless there's a very specific reason - this should be moved to `.css.ts`

packages/react/src/textarea/Textarea.css.ts
radnan: we shouldn't mess with letter spacing except for maybe headings

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: understood bhai. Addressed this using `data-size` attribute

packages/react/src/textarea/Textarea.css.ts
toha993: @JafarSaifullah

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,70 @@
...
+        data-invalid={error}
+        data-resize={resize}
+        {...styles.wrapper()}
+        {...sprinkleProps}
+      >
+        {topSection && <Box>{topSection}</Box>}
+        <Box
+          asChild
+          {...styles.textarea({}, className)}
+          color={disabled ? "fg.disabled" : "fg.default"}
```
toha993, line 54:
```suggestion
          color={disabled ? "fg.disabled" : "fg.default"}
          {...styles.textarea({}, className)}
```

JafarSaifullah: addressed

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,75 @@
...
+export const wrapper = recipe({
+  base: [
+    {
+      border: "1",
+      fontSize: "md",
+      gap: "xs",
+      overflow: "auto",
+      p: "xs",
+      rounded: "sm",
+      w: "340",
```
radnan, line 13: elements should not have defined size - and should take up the container width
```suggestion
```

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <Flex
+        aria-disabled={disabled}
+        aria-invalid={error}
+        data-disabled={disabled}
+        data-invalid={error}
+        data-resize={resize}
+        {...styles.wrapper()}
+        {...sprinkleProps}
+      >
+        {topSection && <Box>{topSection}</Box>}
```
radnan, line 50: let's not wrap the top and bottom sections with box
```suggestion
        {topSection}
```

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,70 @@
...
+          color={disabled ? "fg.disabled" : "fg.default"}
+          {...styles.textarea({}, className)}
+        >
+          <textarea
+            placeholder={placeholder}
+            readOnly={disabled || readOnly}
+            ref={ref}
+            rows={rows}
+            {...restProps}
+          ></textarea>
```
radnan, line 62: this can be changed to self-closing - looks a bit better
```suggestion
          />
```

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </Box>
+        {bottomSection && <Box>{bottomSection}</Box>}
```
radnan, line 64:
```suggestion
        {bottomSection}
```

packages/react/src/textarea/Textarea.tsx
JafarSaifullah: addressed bhai

packages/react/package.json
```diff
@@ -36,6 +36,7 @@
     "@reach/auto-id": "^0.18.0",
     "@vanilla-extract/sprinkles": "^1.6.2",
     "clsx": "^2.1.1",
+    "framer-motion": "^11.2.10",
```
radnan, line 39: don't need this?

apps/storybook/src/forms/Textarea.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Flex alignItems="stretch" justifyContent="center">
+        <Button appearance="primary">Bottom Section</Button>
+      </Flex>
+    ),
+    placeholder: "Write your text",
+    topSection: <Text>Top Section</Text>,
+  },
+};
+
+export const TypesOfTextarea: Story = {
```
radnan, line 66: don't need this story probly - since all the other stories already cover these

apps/storybook/src/forms/Textarea.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export const NoResize: Story = {
+  args: {
+    placeholder: "Enter text....",
+    resize: "none",
+  },
+};
+
+export const Custom: Story = {
```
radnan, line 54: can we use a better name here? or maybe make it more like the opal design - or remove it since the other stories already cover the sections

packages/react/src/textarea/Textarea.tsx
radnan: we're kind of taking over the `resize` prop and extending its functionality. since in the case of auto sizing you don't really want the user to be able to manually control the size - so they're mutually exclusive

apps/storybook/src/forms/Textarea.stories.tsx
JafarSaifullah: contains the `error` states and the story similar to `Input` story.
We can remove it nonetheless

packages/react/package.json
JafarSaifullah: not entirely sure on this part bhai.

apps/storybook/src/forms/Textarea.stories.tsx
JafarSaifullah: gotcha bhai

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -0,0 +1,77 @@
...
+  },
+});
+export const textarea = recipe({
+  base: [
+    {
+      alignItems: "start",
+      flex: "1",
+    },
+    style({
+      borderColor: theme.colors["border.default"],
```
toha993, line 61: Lets move it to sprinkles as we are not updating it in selectors

packages/react/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,70 @@
...
+import * as styles from "./Textarea.css";
+
+type TextareaProps = ExtendProps<
+  ComponentPropsWithRef<"textarea">,
+  ComponentPropsWithRef<typeof Box>,
+  {
+    bottomSection?: ReactNode;
+    disabled?: boolean;
+    error?: boolean;
+    topSection?: ReactNode;
```
toha993, line 17: Let's move with startDecorator and endDecorator for top/bottom.

toha993 (approved):

Issue #49, opened 2024-05-28
add progress bar component


ImperishableMe:
We are using the [radix-ui primitives](https://www.radix-ui.com/primitives/docs/components/progress) for implementing the Progress bar. They expose the following properties:

`value` - Optional. Signifies the current progress. Defaults to `0`.

`max` - optional. Signifies the maximum possible value. Default: `100`.

`getValueLabel` - optional (very unlikely that anyone will change this). A function that determines what will be the [aria-valuetext](https://github.com/radix-ui/primitives/blob/b32a93318cdfce383c2eec095710d35ffbd33a1c/packages/react/progress/src/Progress.tsx#L51C11-L51C38).

ImperishableMe:
@radnan bhaia, I want to start working on this. I will update the issue with suggested API very soon. Is it possible to assign this issue to me?

ImperishableMe:
The design shown [here](https://www.figma.com/design/GH25lYLDM34molSm1yhQO3/Opal-Guidelines?m=dev&node-id=2437-70026) expects a different colour #7C3AED that is not currently in our design system. 
Should I take the colour of the progress bar indicator as a prop?
cc: @radnan bhaia.

PR #58, opened 2024-06-03
Add Progress Bar Component

ImperishableMe:
Issue https://github.com/optimizely-axiom/optiaxiom/issues/49


packages/react/src/progress-bar/ProgressBar.recipe.ts
```diff
@@ -0,0 +1,33 @@
...
+    overflow: "hidden",
+  },
+  variants: {
+    size: {
+      default: { h: "6", w: "1/2" },
+    },
+    variant: {
+      default: {
+        bg: "white",
+      },
```
ImperishableMe, line 15: @radnan bhaia, as I mentioned in the https://github.com/optimizely-axiom/optiaxiom/issues/49#issuecomment-2144331490, the given design for [Opal frontend](https://www.figma.com/design/GH25lYLDM34molSm1yhQO3/Opal-Guidelines?m=dev&node-id=2437-70026), we need another colour in (#7C3AED) to support that.
How can we support that?
Should we add another variant (maybe called `ai`?) to the recipe list? Or expose a prop for the `color`?

apps/storybook/src/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { Flex, ProgressBar } from "@optiaxiom/react";
+import React from "react";
+
+const meta: Meta<typeof ProgressBar> = {
+  component: ProgressBar,
+  title: "Primitives / ProgressBar",
```
mostafa-K-raihan, line 8: I think this should go under `Feedback`
```suggestion
  title: "Feedback / ProgressBar",
```

packages/react/src/progress-bar/ProgressBar.recipe.ts
toha993: I think, if the colour is base for the design, we should add it in our colour variant as semantic tokens, then we use it in our variants rather than exposing colour prop to always pass colour for the component.

apps/storybook/src/ProgressBar.stories.tsx
ImperishableMe: Addressed!

packages/react/src/progress-bar/ProgressBar.recipe.ts
ImperishableMe: Thanks for the suggestion. However, as per the suggestion of @radnan bhaia, we are keeping the `bg.brand.solid` for now. And for the need of Opal frontend, we will modify the `--ax-colors-bg-success`variable providing a css `classname` as a prop to the component.

toha993: As we are getting away from sprinkles, this sounds the right approach vai.

radnan (approved):

Issue #60, opened 2024-06-03
Add Separator Component


SlowDecay:
We are going to use the Radix UI primitive [separator](https://www.radix-ui.com/primitives/docs/components/separator) to implement this component.

Available props:
1. `orientation`: This can take a value of `horizontal` or `vertical`
2. `decorative`: This can take a boolean value. If this is `true`, then the component is purely decorative and has no semantic purpose. It will be removed from the accessibility tree.

SlowDecay:
@radnan bhaia, I want to start working on this. I will update the issue with a summary of the API very soon. Can you please assign this issue to me if it is ok?

PR #65, opened 2024-06-05
Add Separator Component

SlowDecay:
Issue: https://github.com/optimizely-axiom/optiaxiom/issues/60


apps/storybook/src/Separator.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { Flex, Separator, Text } from "@optiaxiom/react";
+
+const meta: Meta<typeof Separator> = {
+  component: Separator,
+  title: "Components / Separator",
```
radnan, line 7: let's move this inside the `data-display` directory and remove this explicit `title` property

packages/react/src/separator/Separator.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const Separator = forwardRef<HTMLHRElement, SeparatorProps>(
+  ({ asChild, ...props }, ref) => {
+    const Comp = asChild ? Slot : RadixSeparator.Root;
+
+    return (
+      <Box asChild ref={ref} {...props} {...recipe(props)}>
+        <Comp />
+      </Box>
+    );
```
radnan, line 23: i don't think we need to use `Slot` here - it's a single empty element so consumers will probly never need to swap it out for something else
```suggestion
    return (
      <Box asChild ref={ref} {...props} {...recipe(props)}>
        <RadixSeparator.Root />
      </Box>
    );
```

apps/storybook/src/Separator.stories.tsx
SlowDecay: Done in https://github.com/optimizely-axiom/optiaxiom/pull/65/commits/f0e78d7fa1c3c4e4d059679a6f6e31fdb7577946.

packages/react/src/separator/Separator.tsx
SlowDecay: Done in https://github.com/optimizely-axiom/optiaxiom/pull/65/commits/7058c9cb762aad5efd9dc79ad3dcaeae06c45a9a.

radnan (approved):

Issue #67, opened 2024-06-06
(Button) a11y improvements


MunifTanjim:
### Disabled State

Disabled button should have `aria-disabled`

Ref: https://css-tricks.com/making-disabled-buttons-more-inclusive/

If button is rendered as anchor, i.e. `<Button asChild><a>Click Me</a></Button>`, the same behavior should be available.

### Keyboard Interaction

Should trigger `onClick` if `Space` or `Enter` is pressed when button is focused.

If button is rendered as anchor, i.e. `<Button asChild><a>Click Me</a></Button>`, the same interaction should be available.

Ref: https://www.w3.org/WAI/ARIA/apg/patterns/button/#:~:text=Keyboard%20Interaction-,When%20the%20button%20has%20focus%3A,%3A%20Activates%20the%20button.,-Following%20button%20activation

PR #69, opened 2024-06-06
fix disabled state interactions

JafarSaifullah:
resolves https://github.com/optimizely-axiom/optiaxiom/issues/67

### DIFF
--- a/packages/react/src/button/Button.css.ts
+++ b/packages/react/src/button/Button.css.ts
@@ -26,7 +26,7 @@ export const button = recipe({
         transitionTimingFunction: "ease",
 
         selectors: {
-          '&:active:not([data-disabled="true"])': {
+          '&:active:not([aria-disabled="true"])': {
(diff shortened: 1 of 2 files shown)


packages/react/src/button/Button.css.ts
```diff
@@ -69,51 +69,51 @@ export const button = recipe({
...
       },
       primary: {
         "@layer": {
           [layers.axiom]: {
-            vars: {
-              [accentColorVar]: theme.colors["bg.brand.solid"],
-              [solidAccentColorVar]: theme.colors["bg.brand.solid.hover"],
-              [subtleAccentColorVar]: theme.colors["bg.brand.subtle"],
-            },
-
```
MunifTanjim, line 94: Can we revert these unnecessary changes?

JafarSaifullah: addressed bhai

MunifTanjim (approved):

Issue #68, opened 2024-06-06
(ProgressBar) style issues


MunifTanjim:
- Width is fixed at `50%`
- No way to override the width easily
- Border radius looks odd, the outer part is squared but inner part is rounded.
    <img width="67" alt="image" src="https://github.com/optimizely-axiom/optiaxiom/assets/8050659/086a5502-17c2-414d-a05e-094fe9db2e2c">

MunifTanjim:
@ImperishableMe Somehow I can't add you to this issue as assignee.

cc: @radnan

ImperishableMe:
Issued https://github.com/optimizely-axiom/optiaxiom/pull/71, but cannot assign you as a reviewer @MunifTanjim bhai.

radnan:
> @ImperishableMe Somehow I can't add you to this issue as assignee.
> 
> cc: @radnan

this is an odd and annoying thing about GH - seems until the user interacts with the issue you can't assign them - not sure how we get around that

ImperishableMe:
> Border radius looks odd, the outer part is squared but inner part is rounded.

@MunifTanjim  bhai, in the [Opal design](https://www.figma.com/design/GH25lYLDM34molSm1yhQO3/Opal-Guidelines?m=dev&node-id=1303-37000), we have `border-radius` only for the inner box. Of all the allowed `rounded` values in the `Box` component, `lg` seemed to come close to the design. Should I remove `border-radius` altogether?
 
<img width="545" alt="image" src="https://github.com/optimizely-axiom/optiaxiom/assets/27671454/b4065bab-70af-4305-9fe3-5bd8c2989a8e">

MunifTanjim:
@ImperishableMe this is what I got from Figma:

<img width="377" alt="image" src="https://github.com/optimizely-axiom/optiaxiom/assets/8050659/e2011989-47a9-49e9-b68f-20605d1d6047">

<img width="173" alt="image" src="https://github.com/optimizely-axiom/optiaxiom/assets/8050659/d4d31d21-709e-4d54-a52a-afcb8fa8d1e7">

The whole thing is rounded.

ImperishableMe:
@MunifTanjim bhaia, https://github.com/optimizely-axiom/optiaxiom/pull/71 is merged now, please let me know if that solves all the concern of this issue.

MunifTanjim:
Yes. looks good.

PR #71, opened 2024-06-06
Allow passing Box Props to Progress Bar Component

ImperishableMe:
Issue: #68 

Allows the `Box` Component's props to be passed to the Progress Bar Component.

### DIFF
--- a/packages/react/src/progress-bar/ProgressBar.tsx
+++ b/packages/react/src/progress-bar/ProgressBar.tsx
@@ -13,24 +13,32 @@ export const ProgressBar = forwardRef<
   ElementRef<typeof ProgressPrimitive.Root>,
   ProgressBarProps
 >((props, ref) => {
-  const widthPercentage = ((props.value ?? 0) / (props.max ?? 100)) * 100;
+  const widthPercentage =
+    ((props.value ?? 0) / (props.max ?? DEFAULT_MAX)) * 100;
+  const isValidValue =
+    props.value &&
+    props.value >= 0 &&
+    props.value <= (props.max ?? DEFAULT_MAX);
 
   return (
     <Box asChild h="6" rounded="lg" {...props}>
       <ProgressPrimitive.Root ref={ref}>
-        <Box
-          asChild
-          bg="bg.brand.solid"
-          h="full"
-          rounded="lg"
-          style={{ width: `${widthPercentage}%` }}
-          transition="all"
-        >
-          <ProgressPrimitive.Indicator />
-        </Box>
+        {isValidValue && (
+          <Box
+            asChild
+            bg="bg.brand.solid"
+            h="full"
+            rounded="lg"
+            style={{ width: `${widthPercentage}%` }}
+            transition="all"
+          >
+            <ProgressPrimitive.Indicator />
+          </Box>
+        )}
       </ProgressPrimitive.Root>
     </Box>
   );
 });
 
+const DEFAULT_MAX = 100;
 ProgressBar.displayName = "@optiaxiom/react/ProgressBar";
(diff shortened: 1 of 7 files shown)


packages/react/src/progress-bar/ProgressBar.tsx
```diff
@@ -2,23 +2,27 @@ import * as ProgressPrimitive from "@radix-ui/react-progress";
...
 export const ProgressBar = forwardRef<
   ElementRef<typeof ProgressPrimitive.Root>,
   ProgressBarProps
->((props, ref) => {
+>(({ w = "2/3", ...props }, ref) => {
   const widthPercentage = ((props.value ?? 0) / (props.max ?? 100)) * 100;
 
   return (
-    <Box asChild border="1" h="6" overflow="hidden" w="1/2">
+    <Box asChild border="1" h="6" overflow="hidden" w={w}>
```
radnan, line 19: This shouldn’t be just `w` but we should forward/spread all props here

ImperishableMe: Bhai, just to clarify, by forwarding all props, you mean I should forward all the props that `Box` accepts to the `ProgressBar.Root`?

radnan: yep
```suggestion
    <Box asChild border="1" h="6" {...props}>
```

ImperishableMe: Addressed in [3b33ac6](https://github.com/optimizely-axiom/optiaxiom/pull/71/commits/3b33ac6a39f747ac072cebdeacdafb2cff6c7c93). Kept `h=6`, `overflow=hidden` (otherwise it will overflow if `value` > `max`. Alternatively, I can handle it in JS) as default, but user will be able to override those. Does that look okay bhaia?

radnan: why would `value` be bigger than `max`?

radix throws a prop warning in such cases:

Screenshot 2024-06-06 at 12 54 31 PM

sorry i didn't notice this in the initial PR - but we should not show an indicator at all if the `value` exceeds `max`

it's fine to leave in `overflow: hidden` but we should not rely on it to cap the indicator - since it leads to weird issues with the border-radius:

### proper 100% state
Screenshot 2024-06-06 at 12 58 46 PM

### overflow hidden with value > max
Screenshot 2024-06-06 at 12 58 40 PM

packages/react/src/progress-bar/ProgressBar.tsx
```diff
@@ -2,8 +2,12 @@ import * as ProgressPrimitive from "@radix-ui/react-progress";
...
 import { Box } from "../box";
+import { type ExtendProps } from "../utils";
 
-type ProgressBarProps = ComponentPropsWithRef<typeof ProgressPrimitive.Root>;
+type ProgressBarProps = ExtendProps<
+  ComponentPropsWithRef<typeof ProgressPrimitive.Root>,
+  ComponentPropsWithRef<typeof Box>
+>;
 
 export const ProgressBar = forwardRef<
```
radnan, line 12: sorry but can we rename this component to just `Progress` ?

that way it'll be consistent with the radix primitive - also that's what it's called in majority of other component libraries

packages/react/src/progress-bar/ProgressBar.tsx
```diff
@@ -12,13 +16,13 @@ export const ProgressBar = forwardRef<
   const widthPercentage = ((props.value ?? 0) / (props.max ?? 100)) * 100;
 
   return (
-    <Box asChild border="1" h="6" overflow="hidden" w="1/2">
+    <Box asChild border="1" h="6" overflow="hidden" {...props}>
```
MunifTanjim, line 19: @ImperishableMe this is what I got from Figma:

image

image

The whole thing is rounded.

MunifTanjim: And there's no border. Did we intentionally decide to add border?

cc: @radnan

radnan (approved):

PR #96, opened 2024-06-13
Fix `Button` as per specs

JafarSaifullah:
resolves: https://github.com/optimizely-axiom/optiaxiom/issues/34

Summary of Changes:

1. Rename `preset` to `appearance`: Updated the Button component to use the term `appearance` instead of `preset` for better clarity and consistency.
2. Icon Styling Added
3. Spacing and Gaps Fixed
4. Refactored leftSection & rightSection:
    i. The `leftSection` and `rightSection` properties have been refactored to `icon` and `iconPosition`.
    ii. The new structure does not support having icons on both sides simultaneously, simplifying the component design.

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -10,6 +10,7 @@
   },
   "changesets": [
     "cyan-dogs-repair",
+    "hip-snails-film",
     "hot-maps-reflect",
     "moody-apples-compare",
     "pink-lizards-flash",
--- a/packages/react/CHANGELOG.md
+++ b/packages/react/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @optiaxiom/react
 
+## 0.1.0-next.7
+
+### Patch Changes
+
+- 61a8dac: refactored button appearance and icon props
+
 ## 0.1.0-next.6
 
 ### Patch Changes
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -6,7 +6,7 @@
     "url": "https://github.com/optimizely-axiom/optiaxiom.git"
   },
   "type": "module",
-  "version": "0.1.0-next.6",
+  "version": "0.1.0-next.7",
   "files": [
     "dist/**",
     "LICENSE"


packages/react/src/button/Button.css.ts
```diff
@@ -6,16 +6,43 @@ const accentColorVar = createVar();
 const solidAccentColorVar = createVar();
 const subtleAccentColorVar = createVar();
 
-export const button = recipe({
+export const componentWrapper = recipe({
```
toha993, line 9: we can keep it just as wrapper

packages/react/src/button/Button.css.ts
```diff
@@ -121,7 +129,6 @@ export const button = recipe({
             backgroundColor: subtleAccentColorVar,
           },
           '&[aria-disabled="true"]': {
```
toha993, line 131: We will use data attributes in styling,
`&[data-disabled="true"]`

packages/react/src/button/Button.tsx
```diff
@@ -70,9 +73,27 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
     const finalVariant = variant ?? presetProps.variant;
+    const content = icon ? (
+      <Box className={clsx(styles.icon({ size: size }))}>{icon}</Box>
+    ) : (
+      <Box
+        asChild
+        className={clsx(
+          styles.componentWrapper({
+            size: size,
+          }),
```
toha993, line 87: As we are not merging any class name, we can just use `className={ styles.componentWrapper({
            size,
          })}`

packages/react/src/button/Button.tsx
```diff
@@ -70,9 +73,27 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+        )}
+      >
+        <Box className={clsx(styles.leftSection({ size }))}>{leftSection}</Box>
```
toha993, line 90: We can also remove `clsx` from here

packages/react/src/button/Button.tsx
```diff
@@ -70,9 +73,27 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+        <Slottable>{children}</Slottable>
+        <Box className={clsx(styles.rightSection({ size }))}>
```
toha993, line 92: We can also remove `clsx` from here

packages/react/src/button/Button.css.ts
JafarSaifullah: I think we should keep it as it is. We need aria-disabled for keyboard interactions.

radnan: we'll keep both aria-disabled and data-disabled on the element - but only use `data-*` when styling - since that is the contract - whereas `aria-*` attributes are just a side effect of the intent

JafarSaifullah: understood bhai!

packages/react/src/button/Button.tsx
```diff
@@ -70,9 +73,23 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         </Text>
       );
 
-    const presetProps = presets[preset];
+    const presetProps = appearances[appearance];
     const finalColorScheme = colorScheme ?? presetProps.colorScheme;
     const finalVariant = variant ?? presetProps.variant;
+    const content = icon ? (
+      <Box className={styles.icon({ size: size })}>{icon}</Box>
+    ) : (
```
MunifTanjim, line 81: I don't like this API. It's the evil branching creeping into the component.

Now `<Button>` has 3 extension point `leftSection`, `rightSection` and `icon`. And they can't all work together... It now has if-you-use-this-you-can-use-that issue. I'd much rather prefer `<IconButton>` component or something else that clearly separates it from the `<Button>`.

Just my 2 cents...

cc: @radnan

radnan: I don’t think this branching is necessary 

The purpose of the `icon` prop is to simply set the shape of the button - you can call it `square` in your head if that helps

But it has no relation to what the actual contents of the buttons are

radnan: The `icon` prop is only meant to modify the padding of the button similar to how `size` does

MunifTanjim: So it's a `boolean`? Still I'd need to manually pass it when I'm passing an icon as `children` right? And also remember to set `aria-label`?

radnan: This is why I think having a single `icon` prop would be better for passing the icon rather than separate leftSection/rightSection props

And then we can make `aria-label` required if a ReactElement is passed as `icon` but no `children` is given.

packages/react/src/tokens/colors.ts
```diff
@@ -125,7 +125,7 @@ export const colors = {
   "bg.brand.solid.hover": colorPalette["brand.600"],
   "bg.brand.subtle": colorPalette["brand.50"],
   "bg.default.hover": colorPalette["neutral.75"],
-  "bg.disabled": colorPalette["neutral.50"],
+  "bg.disabled": colorPalette["neutral.150"],
```
radnan, line 128: where did you see this change in figma?

packages/react/src/button/Button.tsx
```diff
@@ -70,10 +71,27 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         </Text>
       );
 
-    const presetProps = presets[preset];
+    const presetProps = appearances[appearance];
     const finalColorScheme = colorScheme ?? presetProps.colorScheme;
     const finalVariant = variant ?? presetProps.variant;
 
+    const content = (
+      <Box
```
radnan, line 79: why do we need this extra box? can't we merge the styles of `wrapper` and `button` ?

packages/react/src/tokens/colors.ts
JafarSaifullah: My bad on this one bhai. I may have seen wrong

JafarSaifullah: reverting the changes

packages/react/src/button/Button.tsx
JafarSaifullah: addressed bhai

packages/react/src/button/Button.tsx
JafarSaifullah: I addressed the `icon` and `iconPosition` part. How to address the `aria-label` part bhai?

packages/react/src/button/Button.tsx
```diff
@@ -48,11 +55,60 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+          <Text
+            as="span"
+            fontSize="inherit"
+            style={{
+              paddingInline: "4px",
+            }}
+          >
+            {children}
+          </Text>
+        );
```
radnan, line 91: is it possible to avoid this `children` wrapping? this won't work in all cases and its best if we can just avoid it completely

packages/react/src/button/Button.css.ts
```diff
@@ -80,22 +79,28 @@ export const button = recipe({
         },
       }),
     },
+    icon: {
+      false: {},
+      true: {},
+    },
     size: {
-      sm: {
+      sm: style({
```
radnan, line 87: we can't use `style` and mix sprinkle props here - i'm not sure we need `style` here at all since everything should be covered under sprinkles

JafarSaifullah: addressed this bhai

packages/react/src/button/Button.tsx
JafarSaifullah: Should I completely remove this bhai?
It is affecting the `Link` story. Other than that, everything else looks the same

JafarSaifullah: removed the `children` wrapping.

PR #100, opened 2024-06-20
add `RadioGroup` component

toha993:
Add radio component design and story


packages/react/src/radio/Radio.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ComponentPropsWithRef<typeof Box>,
+  {
+    defaultValue?: string;
+    label: string;
+    options: optionProps[];
+    readonly?: boolean;
+  }
+>;
+
+export const Radio = forwardRef<HTMLDivElement, RadioProps>(
```
radnan, line 30: let's call this `RadioGroup` same as the radix primitive

packages/react/src/radio/Radio.css.ts
```diff
@@ -0,0 +1,70 @@
...
+  flexDirection: "row",
+  gap: "8px",
+  selectors: {
+    ["&:not(:first-child)"]: {
+      marginTop: "12px",
+    },
+  },
+});
+
+export const RadioGroupItem = style({
```
radnan, line 15: styles should not be `PascalCased` and should be `camelCased` instead - and always preferably a single word
```suggestion
export const item = style({
```

packages/react/src/radio/Radio.css.ts
```diff
@@ -0,0 +1,70 @@
+import { theme } from "../styles";
+import { style } from "../vanilla-extract";
+
+export const optionWrapper = style({
```
radnan, line 4: can be simplified to just option
```suggestion
export const option = style({
```

packages/react/src/radio/Radio.css.ts
```diff
@@ -0,0 +1,70 @@
...
+  alignItems: "center",
+  flexDirection: "row",
+  gap: "8px",
+  selectors: {
+    ["&:not(:first-child)"]: {
+      marginTop: "12px",
+    },
```
radnan, line 11: why do we need to do this? can we use `gap` instead of margin

packages/react/src/radio/Radio.css.ts
```diff
@@ -0,0 +1,70 @@
...
+    "&[data-disabled]": {
+      borderColor: theme.colors["border.secondary"],
+    },
+    "&[data-state='checked']&:not([data-disabled])": {
+      borderColor: theme.colors["brand.500"],
+    },
+  },
+});
+
+export const RadioGroupIndicator = style({
```
radnan, line 44: name can be simplified to 
```suggestion
export const indicator = style({
```

packages/react/src/radio/Radio.css.ts
toha993: Yes Vai we can. Instead of using box, I used Flex in the updated implementation.

packages/react/src/radioGroup/RadioGroup.css.ts
```diff
@@ -0,0 +1,71 @@
...
+import { style } from "../vanilla-extract";
+
+export const option = style({
+  alignItems: "flex-start",
+  flexDirection: "row",
+  gap: "8px",
+});
+
+export const item = style({
+  all: "unset",
```
MunifTanjim, line 11: @toha993 this is setting the `box-sizing` to `content-box`. That is why you were seeing weird sizing issue (border being calculated outside the width/height).

We shouldn't do this. `box-sizing` should remain set to `border-box`.

toha993: Got it vai.

packages/react/src/radioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,75 @@
...
+import * as RadixRadio from "@radix-ui/react-radio-group";
+import { type ComponentPropsWithRef, forwardRef } from "react";
+
+import type { ExtendProps } from "../utils";
+
+import { Flex } from "../flex";
+import { Text } from "../text";
+import * as styles from "./RadioGroup.css";
+
+type optionProps = {
```
radnan, line 11:
```suggestion
type OptionProps = {
```

type names should always be `PascalCase`

packages/react/src/index.ts
```diff
@@ -10,6 +10,7 @@ export * from "./input";
 export * from "./kbd";
 export * from "./paper";
 export * from "./progress";
+export * from "./radioGroup";
```
radnan, line 13: let's use `kebab-case` for directories

```suggestion
export * from "./radio-group";
```

apps/storybook/src/forms/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,86 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { RadioGroup } from "@optiaxiom/react";
+
+const meta: Meta<typeof RadioGroup> = {
+  component: RadioGroup,
+  parameters: {
+    layout: "centered",
+  },
```
radnan, line 9: we don't need this anymore

packages/react/src/radioGroup/RadioGroup.tsx
radnan: let's add lint rule

packages/react/src/radioGroup/RadioGroup.css.ts
```diff
@@ -0,0 +1,68 @@
+import { theme } from "../styles";
+import { style } from "../vanilla-extract";
+
+export const option = style({
```
radnan, line 4: todo(radnan): add lint rule to check for sprinkles

packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Flex } from "../flex";
+import { RadioGroupItem } from "../radio-group-item";
+import { extractSprinkles } from "../sprinkles";
+
+type RadioGroupProps = BoxProps<
+  typeof RadixRadio.RadioGroup,
+  {
+    children:
+      | ReactElement<typeof RadioGroupItem>
+      | ReactElement<typeof RadioGroupItem>[];
```
radnan, line 22: i don't think this is needed - we can just use
```suggestion
    children?: ReactNode;
```

packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+              )
+            ) {
+              return cloneElement(child, {
+                ...child.props,
+                disabled: child.props?.disabled || disabled,
+                readonly: readonly,
+              });
+            }
+            return child;
+          })}
```
radnan, line 51: we shouldn't do this - there is no guarantee that the children will be available in a flat manner like this - they could be wrapped in fragments - they might be nested deeper inside other components

instead we can re-use the state that radix exposes inside each item and just render the children as-is
```suggestion
          {children}
```

radnan (approved):

PR #101, opened 2024-06-22
add web components output target

radnan:
add package that simply wraps our react components into native web components using preact as a tiny renderer

usage remains largely the same - except elements are prefixed with `ax-` - and slots (props that are react nodes) are handled in the native web component way with `slot` attributes

todo:
- [ ] add rest of the components (figure out a way to make sure we don't miss adding components here

### DIFF
--- a/packages/web-components/README.md
+++ b/packages/web-components/README.md
@@ -1,0 +1,95 @@
+# Axiom Web Components
+
+A framework agnostic variant of the Optimizely Design System.
+
+## Getting Started
+
+```sh
+npm install @optiaxiom/web-components
+```
+
+```tsx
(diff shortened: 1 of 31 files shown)


packages/web-components/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```tsx
+import { Button } from "@optiaxiom/web-components";
+
+function App() {
+  return <Button>Hello World!</Button>;
+}
+```
+
+Please read the full [documentation](https://optimizely-axiom.github.io/optiaxiom/) for guides, examples, and API.
```
radnan, line 19: maybe mention the component naming pattern - `ax-component-name` in kebab case

PR #102, opened 2024-06-23
add `Checkbox` Component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/88


packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+      <Flex
+        aria-disabled={disabled || readonly}
+        data-disabled={disabled || readonly}
+        flexDirection="row"
+        gap="8"
+        ref={ref}
+        {...props}
+      >
+        <Box
+          style={{
```
toha993, line 52: we should use inline `sprinkles` or style in `css.ts`

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <Text
+              color={disabled ? "fg.disabled" : "fg.secondary"}
+              fontSize="sm"
+              style={{
+                letterSpacing: "0.01px",
+                lineHeight: "16px",
+              }}
+            >
+              {helperText}
+            </Text>
```
toha993, line 96: We won't expose helper text, rather exposing label as react node, so devs can pass their own styling and description.

Something like this way and needs to adjust styling according to figma
```suggestion
                     <Text
                asChild
                color={disabled || item.disabled ? "fg.disabled" : "fg.default"}
              >
                <RadixLabel.Root htmlFor={item.value}>
                  <Text asChild {...styles.label}>
                    {item.label}
                  </Text>
                </RadixLabel.Root>
              </Text>
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+              </RadixCheckbox.Indicator>
+            </RadixCheckbox.Root>
+          </Box>
+        </Box>
+
+        <RadixLabel.Root htmlFor={checkboxId} {...styles.rightSection()}>
+          <Text
+            color={disabled ? "fg.disabled" : "fg.default"}
+            fontSize="md"
+            style={{
```
toha993, line 79: We should add the styling in `css.ts`

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,111 @@
...
+});
+
+export const indicator = recipe({
+  base: [
+    style({
+      alignItems: "center",
+      background: theme.colors["bg.brand.solid"],
+      borderRadius: theme.borderRadius["xs"],
+      color: "black",
+      display: "flex",
```
toha993, line 51: We can use sprinkles if available where we aren't overriding them in selectors. 
Keeping `backgroundColor` in style as we are overriding in selectors.

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,111 @@
...
+          cursor: "not-allowed",
+        },
+        '&[data-state="checked"] ': {
+          border: "none",
+        },
+        '&[data-state="indeterminate"]': {
+          border: "none",
+        },
+      },
+      width: theme.spacing["16"],
```
toha993, line 95: Let's move width before selectors and add an empty line under it, so lint doesn't move selectors up.

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,111 @@
...
+        },
+      },
+    }),
+  ],
+});
+
+export const indicatorRoot = recipe({
+  base: [
+    style({
+      border: `1px solid ${theme.colors["neutral.500"]}`,
```
toha993, line 69: Let's use sprinkles whereas we can

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,111 @@
...
+      borderRadius: theme.borderRadius["xs"],
+      color: theme.colors.black,
+      height: theme.spacing["16"],
+      justifyContent: "center",
+      selectors: {
+        "&:focus-visible": {
+          outlineColor: theme.colors["brand.300"],
+          outlineOffset: "1px",
+          outlineStyle: "solid",
+          outlineWidth: "2px",
```
toha993, line 79:
```suggestion
          outline: `2px solid ${theme.colors["brand.300"]}`,
          outlineOffset: "1px",
```

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,111 @@
...
+        '&[data-state="indeterminate"]': {
+          border: "none",
+        },
+      },
+      width: theme.spacing["16"],
+    }),
+  ],
+});
+
+export const rightSection = recipe({
```
toha993, line 100: As it's the styling of label, we can rename it to `label`

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+    label: string;
+    readonly?: boolean;
+  }
+>;
+
+export const Checkbox = forwardRef<HTMLDivElement, CheckboxProps>(
+  (
+    {
+      checked,
+      className,
```
toha993, line 28: Where are we using this `className`

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+        >
+          <Box asChild {...styles.indicatorRoot()}>
+            <RadixCheckbox.Root
+              checked={checked}
+              data-disabled={disabled || readonly}
+              defaultChecked={defaultChecked}
+              disabled={disabled || readonly}
+              id={checkboxId}
+            >
+              <RadixCheckbox.Indicator {...styles.indicator()}>
```
toha993, line 66: We need to wrap it with `box` or `flex` if we want to append sprinkle styling

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+import * as styles from "./Checkbox.css";
+
+type CheckboxProps = ExtendProps<
+  ComponentPropsWithRef<typeof Box>,
+  {
+    checked?: "indeterminate" | boolean;
+    defaultChecked?: "indeterminate" | boolean;
+    disabled?: boolean;
+    helperText?: string;
+    label: string;
```
toha993, line 19:
```suggestion
    label: ReactNode;
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+>;
+
+export const Checkbox = forwardRef<HTMLDivElement, CheckboxProps>(
+  (
+    {
+      checked,
+      className,
+      defaultChecked,
+      disabled,
+      helperText,
```
toha993, line 31:
```suggestion
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+          style={{
+            alignSelf: "start",
+            display: "flex",
+            paddingTop: "4px",
+          }}
+        >
+          <Box asChild {...styles.indicatorRoot()}>
+            <RadixCheckbox.Root
+              checked={checked}
+              data-disabled={disabled || readonly}
```
toha993, line 61: We don't need it here, as Radix gives us `data-disabled` state itself, we should use that rather imposing ours.

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+    },
+    ref,
+  ) => {
+    const checkboxId =
+      id || `checkbox-${label.replace(/\s+/g, "-").toLowerCase()}`;
+
+    return (
+      <Flex
+        aria-disabled={disabled || readonly}
+        data-disabled={disabled || readonly}
```
toha993, line 45: Using the Radix given `disabled` state, we can't make the keyboard interaction, if we want to make the component focusable through the keyboard, we can't use Radix inbuilt `disabled` . So here `aria-disabled` does nothing in fact.

packages/react/src/checkbox/Checkbox.tsx
JafarSaifullah: addressed

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,158 @@
...
+              <RadixCheckbox.Indicator {...styles.indicator()}>
+                {checkboxIcon.indeterminate}
+                {checkboxIcon.checked}
+                {checkboxIcon.unchecked}
+              </RadixCheckbox.Indicator>
+            </RadixCheckbox.Root>
+          </Box>
+        </Box>
+
+        <RadixLabel.Root htmlFor={checkboxId} {...styles.rightSection()}>
```
toha993, line 75: Need to wrap `RadixLabel.Root` with `Box` or `Flex` if we want to add style with sprinkles.

packages/react/src/checkbox/Checkbox.css.ts
JafarSaifullah: fixed this by using sprinkles `h` and `w`

packages/react/src/checkbox/Checkbox.tsx
JafarSaifullah: Addressed it

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,128 @@
...
+  ],
+});
+
+export const indicator = recipe({
+  base: [
+    {
+      rounded: "xs",
+    },
+    style({
+      alignItems: "center",
```
toha993, line 59: Let's use sprinkles for it

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,128 @@
...
+      background: theme.colors["bg.brand.solid"],
+      color: "black",
+      display: "flex",
+      height: "100%",
+      justifyContent: "center",
```
toha993, line 64: Lets use sprinkles for them.

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,128 @@
...
+});
+
+export const rightSection = recipe({
+  base: [
+    {
+      fontSize: "md",
+    },
+    style({
+      cursor: "pointer",
+      letterSpacing: "-.1px",
```
toha993, line 120: Ig, we won't mess up with letterSpacing.

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { Box } from "../box";
+import { Flex } from "../flex";
+import { Text } from "../text";
+import * as styles from "./Checkbox.css";
+
+type CheckboxProps = ExtendProps<
+  ComponentPropsWithRef<typeof Box>,
+  {
+    checked?: "indeterminate" | boolean;
+    children: ReactNode | string;
```
toha993, line 17: Do we need to give `string` type? Can't we pass plain text with ReactNode?

packages/react/src/checkbox/Checkbox.css.ts
JafarSaifullah: based on figma designs. I just copied through

packages/react/src/checkbox/Checkbox.tsx
JafarSaifullah: Unfortunately we can't. So we are wrapping it later with a `Text` component just in case if plain text is passed.

apps/storybook/src/forms/Checkbox.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+export const HelperText: Story = {
+  args: {
+    children: (
+      <Text>
+        Label
+        <Text as="p" color="fg.secondary" fontSize="sm">
+          Helper Text
+        </Text>
+      </Text>
```
radnan, line 26: children is already wrapped in text - so let's use fragment here instead
```suggestion
      <>
        Label
        <Text as="p" color="fg.secondary" fontSize="sm">
          Helper Text
        </Text>
      </>
```

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+  ],
+});
+
+export const iconC = recipe({
+  base: [
+    style({
+      display: "none",
+      selectors: {
+        '[data-state="checked"] &': {
```
radnan, line 19: we shouldn't use unscoped css here - like we've done for `[data-state="checked"]`

let's add a class selector to it like this:

```ts
const wrapperMarker = style({});

export const wrapper = recipe({
  base: [
    {
      alignItems: "start",
      flexDirection: "row",
      gap: "xs",
    },
   wrapperMarker,
  ],
});

export const iconC = recipe({
  base: [
    style({
      display: "none",
      selectors: {
        [`${wrapperMarker}[data-state="checked"] &`]: {
          display: "block",
        },
      },
    }),
  ],
});
```

radnan: we might not have issues with `[data-state="checked"]` - but for something like `[data-disabled="true"]` an ancestor element higher in the tree might have that attribute for whatever reason and will trigger this selector unintentionally

whereas if we use a marker like this then we're guaranteed to only look at attributes on the marker

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const id = useId(idProp);
+
+    return (
+      <Flex
+        data-disabled={disabled || readonly}
+        ref={ref}
+        {...props}
+        {...styles.wrapper({}, className)}
+      >
+        <Flex pt="4">
```
radnan, line 47: let's use the decorator pattern and `align-items: center` instead of hard coded padding here

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <Box asChild {...styles.indicatorWrapper()}>
```
radnan, line 48: i think having both flex and box here is not needed? we can probly collapse these into one element

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,132 @@
...
+      readonly = false,
+      ...props
+    },
+    ref,
+  ) => {
+    const id = useId(idProp);
+
+    return (
+      <Flex
+        data-disabled={disabled || readonly}
```
radnan, line 42: let's switch from a `data-disabled` attribute to a variant for this - since we're controlling this ourselves

`data-disabled` attribute would be needed if we were to expose this state for upstream consumers - or in cases where we're consuming downstream libraries that only expose their state via `data-*` attributes

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,132 @@
...
+        ref={ref}
+        {...props}
+        {...styles.wrapper({}, className)}
```
radnan, line 45: the order here needs to be reversed so we allow consumers to override our own props
```suggestion
        {...styles.wrapper({}, className)}
        {...props}
```

JafarSaifullah: fixed bhai

packages/react/src/checkbox/Checkbox.css.ts
JafarSaifullah: interesting bhai

apps/storybook/src/forms/Checkbox.stories.tsx
jaf107: addressed bhai.
Just exposing endDecorator for this Component

packages/react/src/checkbox/Checkbox.tsx
jaf107: removed the hard-coded padding bhai.

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,89 @@
...
+          </Box>
+
+          <Box asChild {...styles.label()}>
+            <RadixLabel.Root htmlFor={id}>
+              <Text color={disabled ? "fg.disabled" : "fg.default"}>
+                {children}
+              </Text>
+            </RadixLabel.Root>
+          </Box>
+        </Flex>
```
jaf107, line 77: cursor for readonly needs to be discussed

packages/react/src/checkbox/icons-svg/IconChecked.tsx
```diff
@@ -0,0 +1,19 @@
+import * as styles from "../Checkbox.css";
+export const IconChecked = (
```
toha993, line 2: Components should be Pascal Case and everything else should be camelCase

so let's refactor this into a component

```suggestion
export const IconChecked = () => 
```

packages/react/src/checkbox/icons-svg/IconChecked.tsx
```diff
@@ -0,0 +1,19 @@
...
+  <svg
+    {...styles.iconChecked()}
+    fill="none"
```
toha993, line 5: As its need to be controlled by css

```suggestion
    fill="currentColor"
```

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
```diff
@@ -0,0 +1,16 @@
...
+    {...styles.iconIndeterminate()}
+    fill="none"
+    height="8"
+    viewBox="0 0 12 8"
+    width="12"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M2.85714 3.19995C2.35714 3.19995 2 3.56662 2 3.99995C2 4.46662 2.35714 4.79995 2.85714 4.79995H9.14286C9.60714 4.79995 10 4.46662 10 3.99995C10 3.56662 9.60714 3.19995 9.14286 3.19995H2.85714Z"
+      fill="white"
```
toha993, line 13:
```suggestion
      fill="currentColor"
```

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
```diff
@@ -0,0 +1,16 @@
+import * as styles from "../Checkbox.css";
+export const IconIndeterminate = (
```
toha993, line 2:
```suggestion
export const IconIndeterminate = () =>
```

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
```diff
@@ -0,0 +1,16 @@
...
+  <svg
+    {...styles.iconIndeterminate()}
+    fill="none"
```
toha993, line 5: Its not needed here

```suggestion
```

packages/react/src/checkbox/icons-svg/IconUnchecked.tsx
```diff
@@ -0,0 +1,19 @@
+import * as styles from "../Checkbox.css";
+export const IconUnchecked = (
```
toha993, line 2:
```suggestion
export const IconUnchecked = () =>
```

packages/react/src/checkbox/icons-svg/IconUnchecked.tsx
```diff
@@ -0,0 +1,19 @@
...
+  <svg
+    {...styles.iconUnchecked()}
+    fill="none"
```
toha993, line 5:
```suggestion
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+              </Box>
+            </RadixCheckbox.Root>
+          </Box>
+
+          <RadixLabel.Root htmlFor={id}>
+            <Text {...styles.label({ disabled: disabled })}>{children}</Text>
+          </RadixLabel.Root>
+        </Flex>
+        <RadixLabel.Root htmlFor={id}>
+          <Text {...styles.endDecorator({ disabled: disabled })}>
```
toha993, line 85:
```suggestion
          <Text {...styles.endDecorator({ disabled })}>
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  {IconChecked}
+                  {IconUnchecked}
+                  {IconIndeterminate}
+                </RadixCheckbox.Indicator>
+              </Box>
+            </RadixCheckbox.Root>
+          </Box>
+
+          <RadixLabel.Root htmlFor={id}>
+            <Text {...styles.label({ disabled: disabled })}>{children}</Text>
```
toha993, line 81:
```suggestion
            <Text {...styles.label({ disabled })}>{children}</Text>
```

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,168 @@
...
+    },
+    style({
+      borderColor: theme.colors["neutral.500"],
+      selectors: {
+        "&:focus-visible": {
+          outline: `2px solid ${theme.colors["brand.300"]}`,
+          outlineOffset: "1px",
+        },
+        "&:hover": {
+          border: `1px solid ${theme.colors["fg.tertiary"]}`,
```
toha993, line 106:
```suggestion
          borderColor: "fg.tertiary",
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+          </RadixLabel.Root>
+        </Flex>
+        <RadixLabel.Root htmlFor={id}>
+          <Text {...styles.endDecorator({ disabled: disabled })}>
+            {endDecorator}
+          </Text>
+        </RadixLabel.Root>
```
toha993, line 88: Shouldn't it be handled by the end user? Shouldn't we just append the endDecorator here and leave the styling or wrapping to end user?

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
```diff
@@ -0,0 +1,16 @@
+import * as styles from "../Checkbox.css";
```
toha993, line 1: If any other component wants to use the icon, should they have checkbox styling? I think icon should be treated as individual and their styling should be inside the component we want to use. Isn't it?

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+  (
+    {
+      checked,
+      children,
+      className,
+      defaultChecked,
+      disabled,
+      endDecorator,
+      id: idProp,
+      readonly = false,
```
toha993, line 38: we don't need to set falsy values as default

toha993, line 38:
```suggestion
      readonly,
```

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+            disabled: isDisabled,
+          },
+          className,
+        )}
+        {...props}
+      >
+        <Flex flexDirection="row" gap="xs">
+          <Box
+            asChild
+            {...styles.indicatorWrapper({
```
toha993, line 60: I think, `root` makes more sense for this variable name as we are appending the style to root.

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+import * as styles from "./Checkbox.css";
+import { IconChecked } from "./icons-svg/IconChecked";
+import { IconIndeterminate } from "./icons-svg/IconIndeterminate";
+import { IconUnchecked } from "./icons-svg/IconUnchecked";
+
+type CheckboxProps = ExtendProps<
+  ComponentPropsWithRef<typeof Box>,
+  {
+    checked?: "indeterminate" | boolean;
+    children: ReactNode;
```
toha993, line 20: Box already has the children prop. I think we don't need to expose the children prop again.

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,168 @@
...
+});
+
+export const indicatorWrapper = recipe({
+  base: [
+    {
+      border: "1",
+      h: "16",
+      justifyContent: "center",
+      rounded: "xs",
+      w: "16",
```
toha993, line 96:
```suggestion
      size: "16",
      justifyContent: "center",
      rounded: "xs",
```

apps/storybook/src/forms/Checkbox.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+type Story = StoryObj<typeof Checkbox>;
+
+export const Basic: Story = {
+  args: {
+    children: "Label",
+  },
+};
+export const HelperText: Story = {
+  args: {
+    children: <>Label</>,
```
toha993, line 20: Cant we just pass it as string? Do we need the wrapping? 

```suggestion
    children: "Label",
```

apps/storybook/src/forms/Checkbox.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+    endDecorator: "Helper Text",
+  },
+};
+
+export const Disabled: Story = {
+  args: {
+    children: <>Label</>,
```
toha993, line 27:
```suggestion
    children: "Label",
```

apps/storybook/src/forms/Checkbox.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+    disabled: true,
+    endDecorator: "Helper Text",
+  },
+};
+export const Readonly: Story = {
+  args: {
+    children: <>Label</>,
```
toha993, line 34:
```suggestion
    children: "Label",
```

apps/storybook/src/forms/Checkbox.stories.tsx
jaf107: We can. But there are stories in the string format here. So, tried all the possible cases for this

packages/react/src/checkbox/Checkbox.tsx
jaf107: `root` makes more sense to me for the `wrapper` case. But `wrapper` seems better among the two. 
We can rename it to `indicatorRoot` if it seems better. `root` maybe a bit misguiding.

apps/storybook/src/forms/Checkbox.stories.tsx
toha993: Gotcha.

packages/react/src/checkbox/Checkbox.tsx
toha993: wrapper is more like a container, whereas root means the root of the component.

packages/react/src/checkbox/Checkbox.tsx
jaf107: The endDecorator has a definite style as per the designs.

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
jaf107: This part is just handling the rendering. We can leave this part as an exception

packages/react/src/checkbox/Checkbox.tsx
toha993: Should it be within RadixLabel or Text , I am not sure. 

c.c. @radnan

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
toha993: I didn't get the comment.

jaf107: Discuss over a call ?

toha993: I think, we shouldn't have the style in icon, rather than composing the style in checkbox component. Something like 

<Box aschild {...styles.iconIndeterminate({}})}>
   <IconIndeterminate/>
</Box>

packages/react/src/checkbox/Checkbox.tsx
jaf107: We're needing a RadixLabel in order to make sure checkbox gets selected when helper text is pressed

packages/react/src/checkbox/Checkbox.tsx
jaf107: it is a container imo :3

packages/react/src/checkbox/icons-svg/IconIndeterminate.tsx
jaf107: Got it

packages/react/src/checkbox/Checkbox.tsx
toha993: Yeah I know - not sure if it should be like that with wrapping.

packages/react/src/checkbox/Checkbox.tsx
toha993: Naming is hard for sure 👍 but I guess it shouldn't be indicatorWrapper for sure.

packages/react/src/checkbox/Checkbox.tsx
radnan: we can just use `cursor: default` instead of pointer for readonly state

packages/react/src/checkbox/icons-svg/IconUnchecked.tsx
radnan: none can just be removed
```suggestion
```

packages/react/src/checkbox/Checkbox.tsx
radnan: we don't need to apply any styles to the decorators - consumers are free to inject whatever they like here

apps/storybook/src/forms/Checkbox.stories.tsx
radnan: we should have a separate story clearly testing this if that's our intent - something like:

```
export const ReactNodeChildren: Story = {
```

otherwise a future dev won't know why this wasn't a string and might change it unintentionally

apps/storybook/src/forms/Checkbox.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+    endDecorator: <>Helper Text</>,
+    readonly: true,
+  },
+};
```
radnan, line 38: let's not add the `endDecorator` to other stories - and only have a single dedicated story showcasing end decorators

packages/react/src/checkbox/Checkbox.tsx
jaf107: We should handle the marginLeft and fontSize at least bhai. No ?

radnan: Margin left yes, font size no

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,141 @@
...
+export const checkbox = recipe({
+  base: [
+    {
+      gap: "0",
+    },
+    wrapperMarker,
+  ],
+  variants: {
+    disabled: {
+      true: style({
```
toha993, line 14: I think we need to have empty object for false.

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+    }),
+  ],
+  variants: {
+    disabled: {
+      true: style({
+        background: theme.colors["fg.disabled"],
+        cursor: "not-allowed",
+      }),
+    },
```
toha993, line 74: I think we need to have empty object for false.

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,91 @@
...
+                </RadixCheckbox.Indicator>
+              </Box>
+            </RadixCheckbox.Root>
+          </Box>
+
+          <RadixLabel.Root htmlFor={id}>
+            <Text {...styles.label({ disabled, readonly })}>{children}</Text>
+          </RadixLabel.Root>
+        </Flex>
+        <Text ml="24">{endDecorator}</Text>
```
toha993, line 85:
```suggestion
        <Box asChild ml="24">{endDecorator}</Box>
```

toha993, line 85: Need to add a check if there's endDecorator passed or not.

packages/react/src/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,89 @@
...
+                </RadixCheckbox.Indicator>
+              </Box>
+            </RadixCheckbox.Root>
+          </Box>
+
+          <RadixLabel.Root htmlFor={id}>
+            <Text {...styles.label({ disabled, readonly })}>{children}</Text>
+          </RadixLabel.Root>
+        </Flex>
+        {endDecorator && <Box ml="24">{endDecorator}</Box>}
```
toha993, line 83: Wont it be with asChild?

packages/react/src/checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,143 @@
...
+      selectors: {
+        "&:hover": {
+          background: theme.colors["bg.brand.solid.hover"],
+        },
+      },
+    }),
+  ],
+  variants: {
+    disabled: {
+      true: style({
```
toha993, line 72: Missed the false object here

packages/react/src/checkbox/Checkbox.tsx
jaf107: fixed

toha993 (approved):

Issue #87, opened 2024-06-12
Add File Uploader Component


ImperishableMe:
Add a FileUploader Component that can be wrapped around a children to give it drag and drop functionality i.e. it will detect any drag and drop events. 

- On the drag events it will check if there are any files
- On the drop event it will fire a callback, passing it a list of files

We can also provide some customization options such as:
- `dragAndDrop`: a `boolean` that signifies whether the component will allow file drag and drop
- `multiple`: a `boolean` that signifies whether the component will allow multiple files to dragged into the area together. 

For the initial version however, maybe we can skip the customisation and provide the basic functionalities.

ImperishableMe:
@radnan bhaia, can you please assign this to me? 
And also, please let me know if the Issue details look reasonable to you.

radnan:
instead of `dragAndDrop` boolean - we can add an `onDrop` prop. so if a user provides that method that means they would like dnd ability and that method will be called with the dropped files

radnan:
feel free to use a third party app like https://react-dropzone.js.org/ to make things easier

radnan:
Not sure if you've started on this already or not - but another good option is https://ark-ui.com/react/docs/components/file-upload since radix-ui doesn't have it yet

ImperishableMe:
> Not sure if you've started on this already or not - but another good option is https://ark-ui.com/react/docs/components/file-upload since radix-ui doesn't have it yet

I am having a look at this bhai. I was OOO for some days.

PR #113, opened 2024-06-27
`AlertDialog` component

toha993:

PR #114, opened 2024-06-27
Add `FileUpload` Component

ImperishableMe:
Issue: https://github.com/optimizely-axiom/optiaxiom/issues/87
This PR adds support for the FileUpload Component. It uses [ark-UI](https://ark-ui.com/react/docs/components/file-upload#anatomy) as a starting point, and exposes all the props that `ark-UI` exposes. 
The root component is the `FileUpload` one. It exposes all of the ark-UI's FileUpload.Root component's props [ref](https://ark-ui.com/react/docs/components/file-upload#api-reference). The sub components are exposed as individual components. However, they are expected to be used wrapped under the root `FileUpload` component. 

The subcomponents are:

- FileUploadDropzone
- FileUploadLabel
- FileUploadTrigger
- FileUploadContext (to get the information about `acceptedFiles`, `rejectedFiles`)


packages/react/src/file-uploader/FileUploader.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <FileUpload.Root maxFiles={5}>
+      <FileUpload.Label>File Upload</FileUpload.Label>
+      <FileUpload.Dropzone>Drag your file(s) here</FileUpload.Dropzone>
+      <FileUpload.Trigger>Choose file(s)</FileUpload.Trigger>
+      <FileUpload.ItemGroup>
+        <FileUpload.Context>
+          {({ acceptedFiles }) =>
+            acceptedFiles.map((file) => (
+              <FileUpload.Item file={file} key={file.name}>
+                <FileUpload.ItemPreview type="image/*">
```
ImperishableMe, line 14: This is the official code from the ark library [ref](https://ark-ui.com/react/docs/components/file-upload) (the first example). It is giving me `ts` error continuously and I could not fix it.
Here is the error:
```
. dev: src/file-uploader/FileUploader.tsx(11,15): error TS2339: Property 'acceptedFiles' does not exist on type 'UseFileUploadContext'.
. dev: src/file-uploader/FileUploader.tsx(12,32): error TS7006: Parameter 'file' implicitly has an 'any' type.
. dev: src/file-uploader/FileUploader.tsx(13,32): error TS2322: Type '{ children: Element[]; file: any; key: any; }' is not assignable to type 'IntrinsicAttributes & FileUploadItemProps & RefAttributes<HTMLLIElement>'.
. dev:   Property 'file' does not exist on type 'IntrinsicAttributes & FileUploadItemProps & RefAttributes<HTMLLIElement>'.
```
Not sure if the library has strict typescript support or not. Can you help me out here? If this does not work out, I can go back to use `react-dropzone`.

radnan: this is a weird quirk of pnpm - @toha993 faced a similar issue in #113

basically you also need to explicitly add the `@zag-js/file-upload` package

```diff
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -36,6 +36,7 @@
     "@radix-ui/react-tooltip": "^1.0.7",
     "@reach/auto-id": "^0.18.0",
     "@vanilla-extract/sprinkles": "^1.6.2",
+    "@zag-js/file-upload": "^0.58.3",
     "clsx": "^2.1.1",
     "inter-ui": "^4.0.2"
   },
```

packages/react/src/file-uploader/FileUploader.tsx
```diff
@@ -0,0 +1,31 @@
+import { FileUpload } from "@ark-ui/react";
+
+export const FileUploader = () => {
```
radnan, line 3: let's call this `FileUpload` similar to the upstream component
```suggestion
export const FileUpload = () => {
```

packages/react/src/file-uploader/FileUploader.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <FileUpload.Root maxFiles={5}>
+      <FileUpload.Label>File Upload</FileUpload.Label>
+      <FileUpload.Dropzone>Drag your file(s) here</FileUpload.Dropzone>
+      <FileUpload.Trigger>Choose file(s)</FileUpload.Trigger>
+      <FileUpload.ItemGroup>
```
radnan, line 9: let's skip this whole `ItemGroup` concept for now - we can add it in a later iteration

packages/react/src/file-uploader/FileUploader.tsx
```diff
@@ -0,0 +1,31 @@
+import { FileUpload } from "@ark-ui/react";
+
+export const FileUploader = () => {
+  return (
+    <FileUpload.Root maxFiles={5}>
+      <FileUpload.Label>File Upload</FileUpload.Label>
+      <FileUpload.Dropzone>Drag your file(s) here</FileUpload.Dropzone>
+      <FileUpload.Trigger>Choose file(s)</FileUpload.Trigger>
```
radnan, line 8: these should come as children

so let's just re-expose the `FileUpload.Label`, `FileUpload.Dropzone`, and `FileUpload.Trigger` components - so devs can choose to piece them back together how they want

the re-exposed `FileUpload.Dropzone` should have styles that make it take up the whole parent container etc.

packages/react/src/file-uploader/FileUploader.tsx
radnan: hmm i'm finding a lot of weirdness with the ark-ui package

let's switch back to the simpler `react-dropzone` but use ark-ui's API and use a similar one in ours

packages/react/src/file-uploader/FileUploader.tsx
ImperishableMe: Yes, bhaia. I have exposed these components: 
- FileUploadDropzone
- FileUploadLabel
- FileUploadTrigger
- FileUploadContext (to get the information about `acceptedFiles`, `rejectedFiles`)
- FileUpload (the root component)

I kept the design to the minimum. Please let me know if we want to enforce any design from any of these components.

apps/storybook/src/data-display/FileUpload.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+              asChild
+              bg="bg.neutral"
+              border="1"
+              h="full"
+              justifyContent="center"
+              rounded="sm"
+              w="full"
+            >
+              <FileUploadDropzone>Drag your files here</FileUploadDropzone>
+            </Flex>
```
toha993, line 57: We could wrap the components with `flex` or `Box` in export where we would just use the exported component and apply props, if necessary, rather than wrapping them in story.

apps/storybook/src/data-display/FileUpload.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+                      {rejectedFiles.map((fileRejection) => (
+                        <li key={fileRejection.file.name}>
+                          {fileRejection.file.name}
+                        </li>
+                      ))}
+                    </ul>
+                  </Box>
+                </Flex>
+              );
+            }}
```
toha993, line 90: Can we wrap them in a component and pass the `acceptedFiles` or `rejectedFiles` to that component?  I am not sure about the design. but as it's repeating, can we do it? We can give an option to use our style, or they just use their own designs if they want by not importing our component with the certain style.

PR #125, opened 2024-06-30
Fix border for input component

toha993:


packages/react/src/input/Input.css.ts
```diff
@@ -6,13 +6,14 @@ export const wrapper = recipe({
...
       alignItems: "center",
       border: "1",
-      borderColor: "border.default",
       color: "fg.default",
       display: "flex",
       flexDirection: "row",
       rounded: "sm",
     },
     style({
+      borderColor: "border.default",
```
radnan, line 15: Needs to be fetched from `theme`
```suggestion
      borderColor: theme.colors["border.default"],
```

radnan: todo: write a lint rule to catch these

radnan (approved):

PR #126, opened 2024-07-01
`Dialog` component

toha993:
The Dialog component is a flexible and customizable modal interface designed for displaying important information, gathering user input, or requesting confirmation. It provides a structured way to present content that temporarily overlays the main application interface.

Key Features:
1. Modular Structure: Composed of sub-components (DialogContent, DialogTitle, DialogBody, DialogFooter) for easy customization.
2. Controlled State: Uses 'open' prop and 'onOpenChange' callback for precise control over dialog visibility.
3. Outside Interaction Handling: Includes 'onInteractOutside' prop to manage clicks outside the dialog.
4. Scrollable Content: DialogBody accommodates varying amounts of content with automatic scrolling.
5. Customizable Footer: DialogFooter allows for flexible arrangement of action buttons.
6. Accessibility: Built with accessibility in mind, ensuring keyboard navigation and screen reader support.

The component's flexibility allows it to adapt to different use cases while maintaining a consistent look and feel across the application.


packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <RadixDialog.Portal>
+              <Flex asChild {...styles.overlay()}>
+                <RadixDialog.Overlay>
+                  <Box asChild {...styles.content({ size })}>
+                    <RadixDialog.Content aria-describedby={undefined}>
+                      {children}
+                      {!hideCloseIcon && (
+                        <Box asChild {...styles.close()}>
+                          <RadixDialog.Close aria-label="Close" asChild>
+                            {CloseIcon}
```
radnan, line 34: let's use a button here - so users can use keyboard to use it
```suggestion
                            <Button appearance="secondary" icon={CloseIcon} />
```

packages/react/src/dialog/Dialog.css.ts
```diff
@@ -0,0 +1,65 @@
...
+    }),
+  ],
+});
+
+export const close = recipe({
+  base: [
+    {
+      p: "2",
+      rounded: "full",
+      size: "20",
```
radnan, line 51: let's remove the padding and sizing since `<Button />` will take care of that
```suggestion
      rounded: "full",
```

packages/react/src/dialog/Dialog.css.ts
```diff
@@ -0,0 +1,65 @@
...
+      shadow: "md",
+    },
+    style({
+      position: "relative",
+    }),
+  ],
+  variants: {
+    size: {
+      sm: style({
+        maxWidth: "90dvw",
```
radnan, line 18: we can move this to the `base` style since the maxWidth is same for all 3 variants

packages/react/src/dialog-footer/DialogFooter.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { Flex } from "../flex";
+import * as styles from "./DialogFooter.css";
+
+type FooterProps = ExtendProps<
+  ComponentPropsWithRef<typeof Flex>,
+  {
+    children: ReactNode;
+  }
+>;
```
radnan, line 14: we can simplify this
```suggestion
type FooterProps = ComponentPropsWithRef<typeof Flex>;
```

packages/react/src/dialog-footer/DialogFooter.tsx
```diff
@@ -0,0 +1,26 @@
+// import * as RadixDialog from "@radix-ui/react-dialog";
```
radnan, line 1: let's remove this line if not needed

packages/react/src/dialog-footer/DialogFooter.css.ts
```diff
@@ -0,0 +1,17 @@
...
+    {
+      flexDirection: "row",
+      gap: "md",
+      justifyContent: "end",
+      px: "24",
+      py: "20",
+    },
+    style({
+      borderTop: `1px solid ${theme.colors["border.secondary"]}`,
+    }),
```
radnan, line 15: i've added sprinkles for individual borders #154

so we can replace this with 
```suggestion
      borderTop: "1",
    },
```

packages/react/src/dialog-title/DialogTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const DialogTitle = forwardRef<HTMLDivElement, FooterProps>(
+  ({ children, ...props }, ref) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+    return (
+      <Flex
+        asChild
+        fontSize="2xl"
+        fontWeight="600"
+        mb="0"
```
radnan, line 22: don't think we need this?
```suggestion
```

packages/react/src/dialog-title/DialogTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+type FooterProps = ExtendProps<
+  ComponentPropsWithRef<typeof Flex>,
+  ComponentPropsWithRef<typeof RadixDialog.Title>
+>;
+
+export const DialogTitle = forwardRef<HTMLDivElement, FooterProps>(
+  ({ children, ...props }, ref) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+    return (
+      <Flex
```
radnan, line 18: let's use `<Heading />` with the correct level prop here instead of manually setting the font size

packages/react/src/dialog-title/DialogTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+        asChild
+        fontSize="2xl"
+        fontWeight="600"
+        mb="0"
+        pb="16"
+        pt="24"
+        px="24"
```
radnan, line 25: let's use the semantic names here for spacing
```suggestion
        pb="md"
        pt="lg"
        px="lg"
```

packages/react/src/dialog/CloseIcon.tsx
```diff
@@ -0,0 +1,14 @@
+export const CloseIcon = (
```
radnan, line 1: components should be PascalCase and everything else should be `camelCase`

so let's refactor this into a component
```suggestion
export const CloseIcon = () => (
```

packages/react/src/dialog/CloseIcon.tsx
```diff
@@ -0,0 +1,14 @@
...
+  <svg
+    fill="none"
+    height="16"
+    viewBox="0 0 20 16"
+    width="20"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M14.8436 12.875C14.6561 13.0625 14.3124 13.0625 14.1249 12.875L9.99988 8.71875L5.84363 12.875C5.65613 13.0625 5.31238 13.0625 5.12488 12.875C4.93738 12.6875 4.93738 12.3438 5.12488 12.1562L9.28113 8L5.12488 3.875C4.93738 3.6875 4.93738 3.34375 5.12488 3.15625C5.31238 2.96875 5.65613 2.96875 5.84363 3.15625L9.99988 7.3125L14.1249 3.15625C14.3124 2.96875 14.6561 2.96875 14.8436 3.15625C15.0311 3.34375 15.0311 3.6875 14.8436 3.875L10.6874 8L14.8436 12.1562C15.0311 12.3438 15.0311 12.6875 14.8436 12.875Z"
+      fill="#080736"
```
radnan, line 11: fill needs to be controllable from css
```suggestion
      fill="currentColor"
```

packages/react/src/dialog/Dialog.css.ts
```diff
@@ -0,0 +1,65 @@
...
+      size: "20",
+    },
+    style({
+      position: "absolute",
+      right: 24,
+      top: 24,
+
+      selectors: {
+        "&:hover": { backgroundColor: theme.colors["gray.200"] },
+      },
```
radnan, line 60: we won't need this once we switch to an actual `<Button />`

apps/storybook/src/overlays/Dialog.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const Default: Story = {
+  render: () => (
+    <DialogTemplate
```
radnan, line 57: let's forward the args here so we can use storybook controls
```suggestion
  render: (args) => (
    <DialogTemplate
      {...args}
```

apps/storybook/src/overlays/Dialog.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+}) => {
+  const [open, setOpen] = useState(false);
+
+  const handleOpen = () => setOpen(true);
+  const handleClose = () => setOpen(false);
+
+  return (
+    <Flex>
+      <Button onClick={handleOpen}>Open Dialog</Button>
+      <Dialog onOpenChange={handleClose} open={open} size={size}>
```
radnan, line 36: let's accept the storybook args and forward them
```suggestion
  size,
  ...props
}: {
  content: string;
  size: "lg" | "md" | "sm";
}) => {
  const [open, setOpen] = useState(false);

  const handleOpen = () => setOpen(true);
  const handleClose = () => setOpen(false);

  return (
    <Flex>
      <Button onClick={handleOpen}>Open Dialog</Button>
      <Dialog onOpenChange={handleClose} open={open} size={size} {...props}>
```

apps/storybook/src/overlays/Dialog.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const Default: Story = {
+  render: () => (
```
radnan, line 56: let's add a `play` function to this default story and assert interactions are working for the modal

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Flex } from "../flex";
+import { CloseIcon } from "./CloseIcon";
+import * as styles from "./Dialog.css";
+
+type DialogProps = ExtendProps<
+  ComponentPropsWithRef<typeof Box>,
+  ComponentPropsWithRef<typeof RadixDialog.Root>,
+  {
+    children: ReactNode;
+    hideCloseIcon?: boolean;
```
radnan, line 16: let's invert this - change this `withCloseButton` and by default it'll be `false`

so we won't show the close button by default - but users can opt-in to that

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+  } & styles.DialogVariants
+>;
+export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
+  ({ children, hideCloseIcon = false, size = "md", ...props }, ref) => {
```
radnan, line 20: technically you don't need to initialize falsy values
```suggestion
  ({ children, hideCloseIcon, size = "md", ...props }, ref) => {
```

apps/storybook/src/overlays/Dialog.stories.tsx
radnan, line 36: let's rename `onOpenChange` to `onClose`

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <Box asChild ref={ref} {...props}>
+        <RadixDialog.Root>
+          <Box asChild>
+            <RadixDialog.Portal>
+              <Flex asChild {...styles.overlay()}>
+                <RadixDialog.Overlay>
```
radnan, line 27: this structure doesn't seem right - box shouldn't be inside the overlay

overlay and box should be siblings

```tsx
<Portal>
  <Overlay />

  <Content />
</Portal>
```

radnan: also, let's invert the pattern for composing box with radix - so instead of having the flex outside let's have it inside

```suggestion
              <RadixDialog.Overlay asChild>
                <Flex {...styles.overlay()}>
```

apps/storybook/src/overlays/Dialog.stories.tsx
toha993: `onOpenChange` is Radix's own exposed function vai.

radnan: i know - but since we're using this dialog in a fully controlled manner and we're not using the dialog trigger from radix - there is never a scenario where this function will be called with `true`. hence the naming `onClose` is much closer to its functionality now.

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+    { children, onClose, size = "md", withCloseButton = false, ...props },
+    ref,
+  ) => {
+    return (
+      <Box asChild ref={ref} {...props}>
+        <RadixDialog.Root onOpenChange={onClose}>
+          <Box asChild>
+            <RadixDialog.Portal>
+              <RadixDialog.Overlay asChild>
+                <Flex {...styles.overlay()}>
```
radnan, line 33: sorry - actually can we go back to the previous usage with the flex outside? that one is more accurate since the radix component should have priority.

the difference is kind of like saying `interface Overlay extends Flex` rather than `interface Flex extends Overlay`

```suggestion
              <Flex asChild {...styles.overlay()}>
                <RadixDialog.Overlay>
```

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+                  <Paper asChild {...styles.content({ size })}>
+                    <RadixDialog.Content>
```
radnan, line 35: let's add some spacing here to make it a bit more readable
```suggestion
      <Box asChild ref={ref} {...props}>
        <RadixDialog.Root onOpenChange={onClose}>

          <Box asChild>
            <RadixDialog.Portal>

              <Flex {...styles.overlay()}>
                <RadixDialog.Overlay asChild>

                  <Paper asChild {...styles.content({ size })}>
                    <RadixDialog.Content>
```

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
+  (
+    { children, onClose, size = "md", withCloseButton = false, ...props },
+    ref,
+  ) => {
+    return (
+      <Box asChild ref={ref} {...props}>
+        <RadixDialog.Root onOpenChange={onClose}>
+          <Box asChild>
+            <RadixDialog.Portal>
```
radnan, line 31: we don't need a box for the portal - since portal is a virtual element and has no dom nodes
```suggestion
          <RadixDialog.Portal>
```

packages/react/src/dialog/Dialog.tsx
radnan: gah - nvm. prettier doesn't allow this.

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Box } from "../box";
+import { Button } from "../button";
+import { Flex } from "../flex";
+import { Paper } from "../paper";
+import { CloseIcon } from "./CloseIcon";
+import * as styles from "./Dialog.css";
+
+type DialogProps = ExtendProps<
+  ComponentPropsWithRef<typeof Box>,
+  ComponentPropsWithRef<typeof RadixDialog.Root>,
```
radnan, line 15: let's pick out the specific props we want from the root
```suggestion
  Pick<ComponentPropsWithRef<typeof RadixDialog.Root>, "open" | "modal">,
```

packages/react/src/dialog/Dialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+  } & styles.DialogVariants
+>;
+export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
+  (
+    { children, onClose, size = "md", withCloseButton = false, ...props },
+    ref,
+  ) => {
+    return (
+      <Box asChild ref={ref} {...props}>
+        <RadixDialog.Root onOpenChange={onClose}>
```
radnan, line 29: i think root is a virtual element as well - it's just a context provider - so we don't need a box here
```suggestion
      <RadixDialog.Root onOpenChange={onClose} {...props}>
```

and it can't have ref either - so let's forward the ref on to the modal overlay instead

radnan (approved):

PR #128, opened 2024-07-02
Add `info` icon in field component

toha993:


apps/storybook/src/forms/Field.stories.tsx
```diff
@@ -12,6 +12,7 @@ type Story = StoryObj<typeof Field>;
 export const Basic: Story = {
   args: {
     children: <Input placeholder="This is an Input" />,
+    info: "Give a proper input",
```
radnan, line 15: let's add a separate story just for `info` - and only have `label` in the Basic story

packages/react/src/field/Field.tsx
```diff
@@ -51,6 +63,13 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
                   *
                 </Text>
               )}
+              {info && (
+                <Tooltip content={info} side="top">
```
radnan, line 67: we don't need to specify the `side` here do we ?

packages/react/src/field/Field.tsx
```diff
@@ -51,6 +63,13 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
...
+                  <Box asChild h="12" ml="2" w="12">
```
radnan, line 68: we can shorten w/h to `size`
```suggestion
                  <Box asChild ml="2" size="12">
```

packages/react/src/field/InfoIcon.tsx
```diff
@@ -0,0 +1,14 @@
...
+      fill="#080736"
```
radnan, line 11: let's change this to currentColor
```suggestion
      fill="currentColor"
```

so we can control the color via css

packages/react/src/field/InfoIcon.tsx
```diff
@@ -0,0 +1,14 @@
+export const InfoIcon = (
```
radnan, line 1: let's name this `IconInfo` - that way when you type `Icon...` your editor will start to autocomplete all available icons in the project

packages/react/src/field/InfoIcon.tsx
```diff
@@ -0,0 +1,14 @@
...
+  <svg
+    fill="none"
```
radnan, line 3: don't think we need to specify `fill` here

packages/react/src/field/Field.tsx
toha993: the gap is `2` vai, but for the icon, the gap in figma is `4`, that's why it was added vai.

radnan: then we can just change the gap to 4 on the parent and remove this `ml` ?

toha993: Vai, the required asterisk has gap 2 from `label`

radnan: aiyah - the info icon should not be inside the `<label>` - you need to introduce an additional `<div>` element here wrapping the label and the info icon and add the gap there

apps/storybook/src/forms/Field.stories.tsx
radnan: let's remove info from the basic story
```suggestion
```

apps/storybook/src/forms/Field.stories.tsx
```diff
@@ -61,3 +63,11 @@ export const DescriptionAndError: Story = {
     label: "Form label",
   },
 };
+
+export const Info: Story = {
+  args: {
+    children: <></>,
```
radnan, line 69: let's add an `<Input />` here
```suggestion
    children: <Input />,
```

packages/react/src/field/Field.tsx
```diff
@@ -42,17 +48,31 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
...
-          </Text>
+          <Flex flexDirection="row" gap="4">
+            <Text
+              alignItems="center"
+              asChild
+              display="flex"
+              fontWeight="500"
+              gap="2"
+            >
+              <RadixLabel.Root htmlFor={id}>
```
radnan, line 59: let's invert this please - so the radix or external component should always come first and inside it should be our component
```suggestion
            <RadixLabel.Root asChild htmlFor={id}>
              <Text
                alignItems="center"
                display="flex"
                fontWeight="500"
                gap="2"
              >
```

packages/react/src/field/Field.tsx
```diff
@@ -37,17 +43,32 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
...
+                  </Text>
+                )}
+              </RadixLabel.Root>
+            </Text>
+            {info && (
+              <>
+                <Tooltip content={info}>
+                  <IconInfo aria-details="iconInfo"></IconInfo>
+                </Tooltip>
+                <VisuallyHidden.Root id="iconInfo">{info}</VisuallyHidden.Root>
```
radnan, line 68: let's use dynamic ID here instead of static one - to handle multiple fields with info on same page
```suggestion
                  <IconInfo aria-details={infoId}></IconInfo>
                </Tooltip>
                <VisuallyHidden.Root id={infoId}>{info}</VisuallyHidden.Root>
```
```tsx
const infoId = useId();
```

PR #136, opened 2024-07-07
Add badge component

toha993:
Add Badge component for status and label display

- Implement Badge with six types: default, primary, secondary, success, warning, danger
- Add custom CSS styles for type-based appearances


packages/react/src/badge/Badge.css.ts
```diff
@@ -0,0 +1,92 @@
...
+          [subtleColorVar]: theme.colors["yellow.700"],
+        },
+      }),
+    },
+    variant: {
+      accent: style({
+        backgroundColor: accentBackgroundColorVar,
+        color: accentColorVar,
+      }),
+      subtle: style({
```
radnan, line 84: let's use the same prop values for button and badge for `variant` prop

so `accent` should become `solid`

and i like the `subtle` name better - so let's change `ghost` to `subtle` in button

packages/react/src/badge/Badge.tsx
```diff
@@ -0,0 +1,58 @@
...
+  {
+    appearance?: keyof typeof appearances;
+    children: ReactNode;
+  } & styles.BadgeVariants
+>;
+
+export const Badge = forwardRef<HTMLDivElement, BadgeProps>(
+  (
+    {
+      appearance = "neutral",
```
radnan, line 29: let's use the same value as button - so change this to `default`
```suggestion
      appearance = "default",
```

radnan, line 29: i don't think we need an `appearance` prop for badge - this was needed in button because we only support a limited number of combinations for button

but for badge consumers will need to use the separate `colorScheme` and `variant` props to control the look

packages/react/src/badge/Badge.tsx
```diff
@@ -0,0 +1,58 @@
...
+  primary: { colorScheme: "primary", variant: "subtle" },
+  success: { colorScheme: "success", variant: "subtle" },
+  warning: { colorScheme: "warning", variant: "subtle" },
+} satisfies Record<string, styles.BadgeVariants>;
+
+type BadgeProps = ExtendProps<
+  ComponentPropsWithRef<typeof Flex>,
+  {
+    appearance?: keyof typeof appearances;
+    children: ReactNode;
```
radnan, line 22: we don't need to define this `children` - as it's already coming from Flex
```suggestion
```

packages/react/src/badge/Badge.tsx
```diff
@@ -0,0 +1,58 @@
...
+>;
+
+export const Badge = forwardRef<HTMLDivElement, BadgeProps>(
+  (
+    {
+      appearance = "neutral",
+      children,
+      className,
+      colorScheme,
+      variant,
```
radnan, line 33: let's default this to `solid`
```suggestion
      variant = 'solid',
```

packages/react/src/badge/Badge.tsx
```diff
@@ -0,0 +1,58 @@
...
+  } & styles.BadgeVariants
+>;
+
+export const Badge = forwardRef<HTMLDivElement, BadgeProps>(
+  (
+    {
+      appearance = "neutral",
+      children,
+      className,
+      colorScheme,
```
radnan, line 32: let's default this to `neutral`
```suggestion
      colorScheme = "netural",
```

packages/react/src/badge/Badge.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <Flex
+        {...styles.badge(
+          { colorScheme: finalColorScheme, variant: finalVariant },
+          className,
+        )}
+        ref={ref}
+        {...props}
+      >
+        <Text as="span" fontSize="inherit">
```
radnan, line 50: i don't think we need both flex and text here - let's combine the two
```suggestion
      <Text
        as="span"
        display="inline-flex"
        fontSize="sm"
        {...styles.badge(
          { colorScheme: finalColorScheme, variant: finalVariant },
          className,
        )}
        ref={ref}
        {...props}
      >
```

radnan: note the explicit `fontSize` and `inline-flex`

packages/react/src/badge/Badge.css.ts
```diff
@@ -0,0 +1,92 @@
...
+const subtleColorVar = createVar();
+const accentBackgroundColorVar = createVar();
+const accentColorVar = createVar();
+
+export const badge = recipe({
+  base: [
+    {
+      alignItems: "center",
+      fontWeight: "500",
+      justifyContent: "center",
```
radnan, line 19: don't need flex related props
```suggestion
      fontWeight: "500",
```

packages/react/src/badge/Badge.css.ts
```diff
@@ -0,0 +1,92 @@
...
+      px: "8",
+      py: "2",
+      rounded: "sm",
+    },
+    style({
+      textTransform: "capitalize",
```
radnan, line 25: this is a sprinkle prop so we don't need a `style()` call for this - but i don't think we need this at all. consumers are free to use whatever content they want and we shouldn't dictate how it should be capitalized.

apps/storybook/src/data-display/Badge.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ...Variants,
+  args: { colorScheme: "warning" },
+};
+
+export const Neutral: Story = {
+  ...Variants,
+  args: { colorScheme: "neutral" },
+};
+
+export const appearance: Story = {
```
radnan, line 71: let's add another story that shows how badge is meant to be used inline with `<Text />` and `<Heading />` components

packages/react/src/badge/Badge.css.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const badge = recipe({
+  base: [
+    {
+      alignItems: "center",
+      fontWeight: "500",
+      justifyContent: "center",
+      px: "8",
+      py: "2",
+      rounded: "sm",
```
radnan, line 22: radius is `xl` according to figma

packages/react/src/badge/Badge.css.ts
toha993: Needed the justify content centre Vai for centring the content.

packages/react/src/badge/Badge.css.ts
toha993: For badge, added the `md` rounding, for indicator will give the `xl` rounding vai.

PR #137, opened 2024-07-07
Export transition component

toha993:


packages/react/src/index.ts
```diff
@@ -19,3 +18,4 @@ export * from "./styles";
 export * from "./text";
 export * from "./tokens";
 export * from "./tooltip";
+export * from "./transition";
```
MunifTanjim, line 21: Are we sure we want to export it with the current interface? Once it's published it'd be hard to change it.

toha993: Got it vai. Should I Close the PR?

MunifTanjim: cc: @radnan

radnan: we'll need to document it - but i think the API is good for now - so we can merge this

radnan: we should also export everything from `animate-presence` since those two go together

toha993: I think we need to change something in web components, it's getting failed in the build.
C.C @radnan

radnan (approved):

Issue #122, opened 2024-06-28
add `Chip` component


radnan:
should be similar to `Button` in terms of color and variants - but more pill shape than button shape

and by default it is informational - meaning no actions are allowed. but they can optionally be actionable as a whole and be able to have icons to the right.

PR #150, opened 2024-07-11
add `Chip` Component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/122


packages/react/src/chip/Chip.tsx
```diff
@@ -0,0 +1,60 @@
...
+type ChipProps = ExtendProps<
+  ComponentPropsWithRef<typeof RadixToggle.Root>,
+  ComponentPropsWithRef<typeof Box>,
+  {
+    children?: ReactNode;
+    disabled?: boolean;
+    endDecorator?: ReactNode;
+    startDecorator?: ReactNode;
+  } & styles.ChipVariants
+>;
```
radnan, line 18: We can now simplify this to:

```suggestion
type ChipProps = BoxProps<
  typeof RadixToggle.Root,
  {
    endDecorator?: ReactNode;
    startDecorator?: ReactNode;
  } & styles.ChipVariants
>;
```

jaf107: Addressed bhai

packages/react/src/chip/Chip.css.ts
```diff
@@ -0,0 +1,117 @@
+import { theme } from "../styles";
+import { createVar, style } from "../vanilla-extract";
+import { type RecipeVariants, recipe } from "../vanilla-extract";
+
+const accentColorVar = createVar();
+const solidAccentColorVar = createVar();
+const subtleAccentColorVar = createVar();
```
toha993, line 7: Do we need this? As there's not much color or tweaking, I think we don't need var.

packages/react/src/chip/Chip.css.ts
```diff
@@ -0,0 +1,117 @@
+import { theme } from "../styles";
+import { createVar, style } from "../vanilla-extract";
+import { type RecipeVariants, recipe } from "../vanilla-extract";
```
toha993, line 3:
```suggestion
import {  type RecipeVariants, createVar, style, recipe } from "../vanilla-extract";
```

packages/react/src/chip/Chip.css.ts
```diff
@@ -0,0 +1,117 @@
...
+    },
+    style({
+      position: "relative",
+      userSelect: "none",
+
+      selectors: {
+        "&:focus-visible": {
+          outlineOffset: "1px",
+          outlineStyle: "solid",
+          outlineWidth: "2px",
```
toha993, line 29: Dont we need color? 

```suggestion
          outline: "2px solid",
          outlineOffset: "1px",          
```

packages/react/src/chip/Chip.css.ts
```diff
@@ -0,0 +1,117 @@
...
+        },
+        '&[data-state="on"]': {
+          backgroundColor: solidAccentColorVar,
+        },
+      },
+    }),
+  ],
+  variants: {
+    colorScheme: {
```
toha993, line 38: I think we can remove colorScheme as it will be single color.

packages/react/src/chip/IconCross.tsx
```diff
@@ -0,0 +1,13 @@
+export const IconCross = () => (
```
toha993, line 1: Need to forward props.

jaf107: Good catch!

packages/react/src/chip/Chip.css.ts
jaf107: it was being set in the ColorScheme part.

packages/react/src/chip/Chip.css.ts
jaf107: I just wanted to wait for the final design. So kept the implementation like this. Changing this though

packages/react/src/chip/Chip.css.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const chip = recipe({
+  base: [
+    {
+      alignItems: "center",
+      display: "inline-flex",
+      flexDirection: "row",
+      gap: "4",
+      justifyContent: "center",
+      overflow: "hidden",
```
toha993, line 17: Why do we need it?

packages/react/src/chip/Chip.tsx
```diff
@@ -0,0 +1,64 @@
...
+          {
+            colorScheme,
+            size,
+            variant,
+          },
+          className,
+        )}
+        {...sprinkleProps}
+      >
+        <Box asChild {...styles.toggleRoot()}>
```
toha993, line 48: We can merge both boxes as one.

packages/react/src/chip/Chip.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <RadixToggle.Root data-disabled={disabled} ref={ref} {...restProps}>
+          {icon && (
+            <Box asChild {...styles.icon({})}>
+              {icon}
+            </Box>
+          )}
+          {children}
+          {endDecorator ?? (!!onPressedChange && <IconCross />)}
+        </RadixToggle.Root>
+      </Box>
```
mutasim-toha, line 53: We should use `disabled` instead of data attribute

PR #156, opened 2024-07-13
Add 'auto-resize' to `Textarea` component

jaf107:
This PR implements auto-resize functionality for the Textarea component, allowing it to automatically adjust its height based on content. The feature is controlled by the `resize` prop and can be fine-tuned with `maxRows` and `minRows` props.

Changes:
- Added `resize` prop to Textarea component
- Implemented auto-resize logic when `resize` is set to `auto`
- Added `maxRows` prop to limit maximum number of rows
- Added `minRows` prop to set minimum number of rows

### DIFF
--- a/packages/react/src/textarea/Textarea.css.ts
+++ b/packages/react/src/textarea/Textarea.css.ts
@@ -41,7 +41,9 @@ export const wrapper = recipe({
   ],
   variants: {
     resize: {
-      auto: {},
+      auto: style({
+        resize: "vertical",
+      }),
       none: style({
         resize: "none",
       }),
@@ -56,10 +58,10 @@ export const textarea = recipe({
     {
       alignItems: "start",
       borderColor: "border.default",
-      flex: "1",
     },
     style({
       color: theme.colors["fg.default"],
+      flexGrow: "1",
       resize: "none",
       selectors: {
         "&:focus-visible": {
--- a/packages/react/src/textarea/Textarea.tsx
+++ b/packages/react/src/textarea/Textarea.tsx
@@ -9,7 +9,7 @@ import { extractSprinkles } from "../sprinkles";
 import * as styles from "./Textarea.css";
 
 type TextareaProps = ExtendProps<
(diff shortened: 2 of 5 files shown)


packages/react/src/textarea/Textarea.css.ts
```diff
@@ -41,7 +41,9 @@ export const wrapper = recipe({
   ],
   variants: {
     resize: {
-      auto: {},
+      auto: style({
+        resize: "vertical",
```
radnan, line 45: this should be "none"
```suggestion
        resize: "none",
```

radnan: in fact you can move this to the base style (and make this and the `none` variant empty objects) and only in the vertical variant you can leave in the `resize: "vertical"` style

packages/react/src/textarea/Textarea.tsx
```diff
@@ -25,6 +26,7 @@ export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
       disabled,
       endDecorator,
       error,
+      maxRows = 5,
```
radnan, line 29: let's not set a default value for `maxRows` here
```suggestion
```

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -56,10 +58,10 @@ export const textarea = recipe({
     {
       alignItems: "start",
       borderColor: "border.default",
-      flex: "1",
```
radnan, line 59: i think we can keep this and just change it to auto - `flex: "auto"`

packages/react/src/textarea/Textarea.css.ts
```diff
@@ -56,10 +58,10 @@ export const textarea = recipe({
...
     },
     style({
       color: theme.colors["fg.default"],
+      flexGrow: "1",
```
radnan, line 64: we won't need this if we set `flex=auto` as per the above comment

PR #158, opened 2024-07-15
Add `Search` component

toha993:
This PR introduces a new `Search` component, enhancing our UI library with a versatile and user-friendly search input functionality.

## Features
- Controlled input with state management
- Start decorator with search icon
- End decorator with clear button (appears when input has value)
- Forwarded ref for direct input access
- Customizable through props inheritance from Input component

## Implementation Details
- Manages input state internally while still allowing external control
- Handles both controlled and uncontrolled scenarios with `defaultValue`
- Implements clear functionality that updates both internal state and notifies parent components

## Usage Example
```jsx
<Search 
  placeholder="Search..." 
/>

### CODE REVIEW

File: packages/react/src/search/Search.tsx [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+
+import { Box } from "../box";
+import { Button } from "../button";
+import { Input } from "../input";
+import { IconCross } from "./IconCross";
+import { IconSearch } from "./IconSearch";
+
+type SearchProps = ComponentPropsWithRef<typeof Input>;
+
+export const Search = forwardRef<HTMLInputElement, SearchProps>(
```
MunifTanjim, line 11: Does it support controlled usage?

packages/react/src/search/Search.tsx
```diff
@@ -0,0 +1,54 @@
...
+        onChange(event);
+      }
+    };
+
+    const handleClearInput = () => {
+      setInputValue("");
+      if (onChange) {
+        onChange({
+          target: { value: "" },
+        } as React.ChangeEvent<HTMLInputElement>);
```
MunifTanjim, line 27: this would be problematic... somebody's gonna try to access something other than `.target.value` and face runtime error.

toha993: Yes vai, I also thought that.
Should I add a onClear prop rather than using the onChange?

packages/react/src/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const Search = forwardRef<HTMLInputElement, SearchProps>(
+  ({ defaultValue = "", onChange, value: valueProp, ...props }, outerRef) => {
+    const innerRef = useRef<HTMLInputElement>(null);
+    const ref = useComposedRefs(innerRef, outerRef);
+
+    const [value, setValue] = useControllableState({
+      defaultProp: defaultValue,
+      onChange: onChange,
+      prop: valueProp,
+    });
```
radnan, line 32: i don't think this is the right approach - your previous approach was the right one

so there's two ways to go about this - either you expose completely custom `onValueChange` and `onValueClear` props (or call them `onChange` and `onClear` in which case you prevent allowing outsiders to pass in their own `onChange` to the actual `<input />`)

or you keep the existing `onChange` handler and when clearing you simply call it again after manually clearing the value on the input

and as for controlled/uncontrolled - you can just do that based on the `typeof valueProp`
```suggestion
    const [innerValue, setValue] = useState(defaultValue);
    const value = typeof valueProp === "undefined" ? innerValue : valueProp;
```

packages/react/src/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <Button
+              appearance="secondary"
+              icon={<IconCross />}
+              onClick={() => setValue("")}
+              size="sm"
+            />
+          )
+        }
+        onChange={(event) => {
+          setValue(event.target.value);
```
radnan, line 47: we need to call the outside `onChange` here if that prop exists
```suggestion
          onChange?.(event);
          setValue(event.target.value);
```

packages/react/src/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    return (
+      <Input
+        endDecorator={
+          value && (
+            <Button
+              appearance="secondary"
+              icon={<IconCross />}
+              onClick={() => setValue("")}
```
radnan, line 41: here we need to do a little trickery

https://stackoverflow.com/a/46012210

```suggestion
              onClick={() => {
                if (!innerRef.current) {
                  return;
                }

                Object.getOwnPropertyDescriptor(
                  innerRef.current.constructor.prototype,
                  "value",
                )?.set?.call(innerRef.current, "");
                innerRef.current?.dispatchEvent(
                  new Event("change", { bubbles: true }),
                );

                innerRef.current.focus();
              }}
```

don't forget to re-focus the input after we clear the value

radnan: we basically re-trigger react's own `onChange` handler on the input - which allows us to set the value or propagate it up to the parent component

packages/react/src/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
...
+              size="sm"
+            />
+          )
+        }
+        onChange={(event) => {
+          setValue(event.target.value);
+        }}
+        ref={ref}
+        startDecorator={<IconSearch />}
+        value={value}
```
radnan, line 51: let's add a `type=search` here
```suggestion
        type="search"
        value={value}
```

packages/react/src/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+type SearchProps = ExtendProps<
+  ComponentPropsWithRef<typeof Box>,
+  Omit<ComponentPropsWithRef<typeof Input>, "onChange">,
+  {
+    defaultValue?: string;
+    onChange?: (value: string) => void;
+    value?: string;
+  }
+>;
```
radnan, line 21: we can simplify this
```suggestion
type SearchProps = ComponentPropsWithRef<typeof Input>;
```

packages/react/src/search/IconCross.tsx
```diff
@@ -0,0 +1,13 @@
+export const IconCross = () => (
+  <svg
+    height="16"
```
radnan, line 3: this component needs to accept `className` for decorator styling
```suggestion
export const IconSearch = ({ className }: { className?: string }) => (
  <svg
    className={className}
    height="16"
```

packages/react/src/input/Input.tsx
```diff
@@ -7,8 +7,8 @@ import { type ExtendProps } from "../utils";
 import * as styles from "./Input.css";
 
 type InputProps = ExtendProps<
-  ComponentPropsWithRef<"input">,
   ComponentPropsWithRef<typeof Box>,
+  ComponentPropsWithRef<"input">,
```
radnan, line 11: we can revert this change and when rebasing against main

radnan (approved):

PR #176, opened 2024-07-24
Add `Toaster` for showing multiple toasts

toha993:


packages/react/src/toaster/Toaster.css.ts
```diff
@@ -0,0 +1,49 @@
...
+        right: 0,
+        top: 0,
+      }),
+    },
+  },
+});
+
+export type ToastPositionVariants = NonNullable<
+  RecipeVariants<typeof viewPort>
+>;
```
radnan, line 49: using NonNullable is not needed
```suggestion
export type ToastPositionVariants = RecipeVariants<typeof viewPort>;
```

packages/react/src/toaster/Toaster.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const { toasts } = useToast();
+    return (
+      <RadixToastProvider>
+        <Flex asChild ref={ref} {...styles.viewPort({ position })} {...props}>
+          <RadixToastViewport>
+            {toasts.map(({ id, ...props }) => (
+              <Toast key={id} {...props} />
+            ))}
+          </RadixToastViewport>
+        </Flex>
```
radnan, line 26: toasts shouldn't be inside the view port children
```suggestion
        {toasts.map(({ id, ...props }) => (
          <Toast key={id} {...props} />
        ))}
        
        <Flex asChild ref={ref} {...styles.viewPort({ position })} {...props}>
          <RadixToastViewport />
        </Flex>
```

packages/react/src/toaster/Toaster.tsx
```diff
@@ -0,0 +1,32 @@
...
+} from "@radix-ui/react-toast";
+import { type ComponentPropsWithRef, forwardRef } from "react";
+
+import { Flex } from "../flex";
+import { Toast } from "../toast/Toast";
+import * as styles from "./Toaster.css";
+import { useToast } from "./use-toast";
+
+type ToasterProps = ComponentPropsWithRef<typeof RadixToastProvider> &
+  styles.ToastPositionVariants;
```
radnan, line 13: let's use `BoxProps` here and extend Viewport
```suggestion
type ToastProps = BoxProps<
  typeof RadixToast.Viewport,
  {
    children?: ReactNode;
  } & ComponentPropsWithoutRef<typeof RadixToast.ToastProvider> &
    styles.ToastPositionVariants
>;
```

PR #177, opened 2024-07-25
Add `Alert` component

toha993:
## Overview
This PR introduces a new `Alert` component to our design system. The Alert component is designed to display important messages to users in a visually distinctive and contextually appropriate manner.

## Features
- Customizable appearance with different types: info, success, warning, and danger
- Two size variants: medium (380px) and large (640px)
- Optional title for added context
- Closable with a custom close handler
- Accessible, using appropriate ARIA attributes

## Usage Example
```jsx
<Alert
  type="success"
  size="md"
  title="Operation Successful"
  onClose={() => console.log('Alert closed')}
>
  Your changes have been saved successfully.
</Alert>

### CODE REVIEW

File: packages/react/src/alert/Alert.tsx [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const Alert = forwardRef<HTMLDivElement, AlertProps>(
+  (
+    {
+      children,
+      onClose,
+      size = "md",
+      title,
+      type = "info",
```
radnan, line 39: let's rename this to `colorScheme`
```suggestion
      colorScheme = "information",
```

packages/react/src/alert/Alert.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const IconComponent = iconMap.get(type);
+  return IconComponent ? <IconComponent /> : null;
+};
+
+export const Alert = forwardRef<HTMLDivElement, AlertProps>(
+  (
+    {
+      children,
+      onClose,
+      size = "md",
```
radnan, line 37: we don't need this - size will be dynamic relative to the container

so if consumers want to use this a banner they'll just set width to 100% - and everywhere else it'll just take the width of the container

packages/react/src/alert/Alert.tsx
```diff
@@ -0,0 +1,68 @@
...
+      title,
```
radnan, line 38: we can remove this unused prop
```suggestion
```

packages/react/src/alert/Alert.css.ts
```diff
@@ -0,0 +1,116 @@
...
+  ],
+  variants: {
+    size: {
+      md: style({ minWidth: "288px" }),
+      lg: style({ minWidth: "540px" }),
+    },
+  },
+});
+
+export const startDecorator = recipe({
```
radnan, line 106: let's rename this to icon
```suggestion
export const icon = recipe({
```

packages/react/src/alert/Alert.css.ts
```diff
@@ -0,0 +1,116 @@
...
+export const close = recipe({
+  base: [
+    {
+      px: "2",
+      py: "4",
+      rounded: "sm",
+      size: "sm",
+    },
+    style({
+      cursor: "pointer",
```
radnan, line 81: close is a button so we don't need this
```suggestion
```

packages/react/src/alert/Alert.css.ts
```diff
@@ -0,0 +1,116 @@
...
+      gridArea: "close",
+      marginRight: "14px",
```
radnan, line 83: this doesn't feel right - aren't we using padding on the parent element? in which case we shouldn't need this

packages/react/src/alert/Alert.css.ts
```diff
@@ -0,0 +1,116 @@
...
+    },
+  },
+});
+
+export const close = recipe({
+  base: [
+    {
+      px: "2",
+      py: "4",
+      rounded: "sm",
```
radnan, line 77: we shouldn't override the default button padding/shape etc

packages/react/src/alert/Alert.css.ts
```diff
@@ -0,0 +1,116 @@
...
+const lightColorVar = createVar();
+
+export const alert = recipe({
+  base: [
+    {
+      alignItems: "start",
+      flexDirection: "row",
+      gap: "xs",
+      pl: "md",
+      py: "md",
```
radnan, line 19: any reason not to use just `p` here?
```suggestion
      p: "md",
```

packages/react/src/alert/Alert.css.ts
```diff
@@ -0,0 +1,116 @@
...
+      rounded: "md",
+    },
+    style({
+      maxWidth: "90dvw",
+    }),
```
radnan, line 24: we don't need this
```suggestion
```

packages/react/src/alert/Alert.tsx
```diff
@@ -0,0 +1,68 @@
...
+      type = "info",
+      variant = "light",
+      ...props
+    },
+    ref,
+  ) => {
+    return (
+      <Flex ref={ref} {...styles.alert({ size, type, variant })} {...props}>
```
radnan, line 46: need to handle className here
```suggestion
      <Flex ref={ref} {...styles.alert({ size, type, variant }, className)} {...props}>
```

packages/react/src/alert-description/AlertDescription.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Flex } from "../flex";
+
+type AlertDescriptionProps = ComponentPropsWithRef<typeof Flex>;
+
+export const AlertDescription = forwardRef<
+  HTMLDivElement,
+  AlertDescriptionProps
+>(({ children, ...props }, ref) => {
+  return (
+    <Flex fontSize="md" fontWeight="400" ref={ref} {...props}>
```
radnan, line 12: let's use `Box` unless we know the children will need flex

packages/react/src/alert-title/AlertTitle.tsx
```diff
@@ -0,0 +1,17 @@
+import { type ComponentPropsWithRef, forwardRef } from "react";
+
+import { Flex } from "../flex";
+
+type AlertTitleProps = ComponentPropsWithRef<typeof Flex>;
+
+export const AlertTitle = forwardRef<HTMLDivElement, AlertTitleProps>(
+  ({ children, ...props }, ref) => {
+    return (
+      <Flex fontWeight="600" ref={ref} {...props}>
```
radnan, line 10: let's use `Box` unless we know the children will need flex

apps/storybook/src/feedback/Alert.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+          {...args}
+        >
+          {args.children}
+        </Alert>
+      )}
+    </Flex>
+  );
+};
+
+export const Default: Story = {
```
radnan, line 113: let's use `Basic` for the most default story name
```suggestion
export const Basic: Story = {
```

apps/storybook/src/feedback/Alert.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+  args: {
+    children: (
+      <AlertDescription>You can update your email only once</AlertDescription>
+    ),
+  },
+  render: DefaultTemplate,
```
radnan, line 119: do we need this `DefaultTemplate` render in all the stories? let's add it only to demonstrate specific things like onClose

packages/react/src/alert/Alert.css.ts
toha993: In figma, the design is pretty odd, close icon have 14px from border vai, elsewhere it is 16px. so added it individually here.

packages/react/src/alert/Alert.css.ts
toha993: didn't use pr here vai for the close icon margin.

radnan (approved):

PR #193, opened 2024-07-26
refactor toast into separate components

radnan:


packages/react/src/toaster/Toaster.tsx
```diff
@@ -0,0 +1,52 @@
...
+    },
+    ref,
+  ) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
+    return (
+      <RadixToast.ToastProvider
+        duration={duration}
+        label={label}
+        swipeDirection={swipeDirection}
```
radnan, line 39: todo: need to set `swipeDirection` based on `position`

packages/react/src/toast/Toast.tsx
```diff
@@ -35,78 +29,39 @@ const getIcon = (type: string) => {
...
-        </Box>
-      </RadixToast.ToastProvider>
+      <Paper
+        asChild
+        elevation="md"
+        {...styles.root({ type })}
+        {...sprinkleProps}
+      >
+        <RadixToast.Root
+          onOpenChange={onOpenChange}
```
radnan, line 44: todo: add animations

apps/storybook/src/feedback/Toast.stories.tsx
```diff
@@ -1,118 +1,107 @@
...
-  render: ToastTemplate,
 };
 
-export const InteractiveTest: Story = {
-  args: {
-    children: "This is an interactive test toast message",
-    position: "bottom-right",
-    type: "info",
-  },
+export const Interactive: Story = {
```
toha993, line 104: Need to change the `text content` as its different from the default text.

radnan: addressed

PR #202, opened 2024-07-31
add `Pagination` component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/183


packages/react/src/pagination/Pagination.tsx
```diff
@@ -0,0 +1,98 @@
...
+    if (totalPage <= 1) return null;
+    return (
+      <Flex
+        asChild
+        ref={ref}
+        {...styles.wrapper({}, className)}
+        {...sprinkleProps}
+      >
+        <Flex {...restProps}>
+          <ButtonGroup gap="2">
```
radnan, line 64: let's collapse these into two elements instead of 3 - `<nav>` and `<ul>`

basically follow the layout in https://design-system.w3.org/components/pagination.html

packages/react/src/pagination/Pagination.tsx
```diff
@@ -0,0 +1,98 @@
...
+            <Button
```
radnan, line 65: let's use custom elements here instead of re-using `Button`

packages/react/src/pagination/Pagination.tsx
```diff
@@ -0,0 +1,98 @@
...
+              appearance="secondary"
+              disabled={active === 1}
+              icon={<IconAngleLeft />}
+              onClick={() => handlePageSelect(previous())}
+            >
+              Previous
+            </Button>
+            {buttons.map(({ children, value, ...props }, index) => (
```
radnan, line 73: ellipsis items should not be buttons - they should not be focusable

jaf107: ellipsis will be disabled. And as such, they are not focusable bhai

radnan (approved):

PR #213, opened 2024-08-07
add `Breadcrumb` component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/186


packages/react/src/breadcrumb-item/BreadcrumbItem.css.ts
```diff
@@ -0,0 +1,22 @@
...
+import { recipe, style } from "../vanilla-extract";
+
+export const breadcrumbItem = recipe({
+  base: [
+    {
+      alignItems: "center",
+    },
+    style({
+      display: "flex",
+    }),
```
toha993, line 11:
```suggestion
    {
      alignItems: "center",
      display: "flex",
    },
```

packages/react/src/breadcrumb-item/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const BreadcrumbItem: React.FC<BreadcrumbItemProps> = ({
+  followingSeparator = false,
+  href,
+  isEllipsis = false,
+  label,
+  onClick,
+  precedingSeparator = true,
+  separator = ">",
+}) => (
```
toha993, line 25: I think we don't follow this convention. 

```suggestion
export const BreadcrumbItem= ({
  followingSeparator = false,
  href,
  isEllipsis = false,
  label,
  onClick,
  precedingSeparator = true,
  separator = ">",
}: BreadcrumbItemProps ) => (
```

if we need to allow ref, we need to use higher order component.

packages/react/src/breadcrumb/Breadcrumb.css.ts
```diff
@@ -0,0 +1,33 @@
...
+});
+
+export const breadcrumbList = recipe({
+  base: [
+    {
+      alignItems: "center",
+    },
+    style({
+      display: "flex",
+    }),
```
toha993, line 21:
```suggestion
    {
      alignItems: "center",
      display: "flex",
    },
```

packages/react/src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { Box, type BoxProps } from "../box";
+import { extractSprinkles } from "../sprinkles";
+import * as styles from "./Breadcrumb.css";
+import { useBreadcrumbItems } from "./useBreadcrumbItems";
+
+export type BreadcrumbProps = BoxProps<
+  "nav",
+  {
+    children: React.ReactNode;
```
toha993, line 11:
```suggestion
    children: ReactNode;
```
We can remove the redundant `React` and import `ReactNode` from named export.

packages/react/src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,40 @@
...
+    maxItems?: number;
+    separator?: React.ReactNode;
```
toha993, line 13:
```suggestion
    separator?: ReactNode;
```

packages/react/src/breadcrumb/Breadcrumb.tsx
toha993, line 11:
```suggestion
```
We have children in our boxProps.

packages/react/src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+>;
+
+export const Breadcrumb = forwardRef<HTMLElement, BreadcrumbProps>(
+  (
+    { children, className, maxItems = Infinity, separator = ">", ...props },
```
toha993, line 19: Do we really need to set Infinity rather than a number as default?

packages/react/src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const visibleItems = useBreadcrumbItems({
+      children,
+      maxItems,
+      separator,
+    });
+
+    return (
+      <Box asChild {...styles.breadcrumb({}, className)} {...sprinkleProps}>
+        <nav aria-label="Breadcrumb" ref={ref} {...restProps}>
+          <ol {...styles.breadcrumbList()}>{visibleItems}</ol>
```
toha993, line 33: I think we need to use `box` as composition here.
```suggestion
         <Box aschild {...styles.breadcrumbList()}>
             <ol >{visibleItems}</ol>
          </Box>
```

packages/react/src/breadcrumb/useBreadcrumbItems.tsx
```diff
@@ -0,0 +1,93 @@
...
+} from "../breadcrumb-item/BreadcrumbItem";
+import { Menu } from "../menu";
+import { MenuContent } from "../menu-content";
+import { MenuItem } from "../menu-item";
+import { MenuTrigger } from "../menu-trigger";
+import { Text } from "../text";
+import * as styles from "./Breadcrumb.css";
+
+type UseBreadcrumbItemsProps = {
+  children: React.ReactNode;
```
toha993, line 21:
```suggestion
  children: ReactNode;
```

packages/react/src/breadcrumb/useBreadcrumbItems.tsx
```diff
@@ -0,0 +1,93 @@
...
+  maxItems: number;
+  separator: React.ReactNode;
```
toha993, line 23:
```suggestion
  separator: ReactNode;
```

packages/react/src/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,40 @@
...
+>;
+
+export const Breadcrumb = forwardRef<HTMLElement, BreadcrumbProps>(
+  (
+    { children, className, maxItems = Infinity, separator = ">", ...props },
+    ref,
+  ) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
+    const visibleItems = useBreadcrumbItems({
```
toha993, line 24: I think we can skip the need to use `hooks` for it, rather can divide the component with more flexibility. Like
1. Breadcrumb
2. BreadcrumbList
3. BreadcrumbItem
4. BreadcrumbSeparator
5. BreadcrumbEllipsis

something like this. 
https://ui.shadcn.com/docs/components/breadcrumb

c.c @radnan vai

packages/react/src/breadcrumb/Breadcrumb.tsx
jaf107: This needs to be discussed together

packages/react/src/breadcrumb/Breadcrumb.tsx
jaf107: I think we can skip this and stick with the current implementation

radnan (approved):

PR #222, opened 2024-08-11
Add `AvatarGroup` component

toha993:


packages/react/src/avatar/Avatar.tsx
```diff
@@ -34,12 +35,15 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(
       colorScheme = "neutral",
       icon,
       name,
-      size = "md",
+      size: _sizeProp,
```
radnan, line 38:
```suggestion
      size: sizeProp,
```

packages/react/src/avatar-context/AvatarContext.tsx
```diff
@@ -0,0 +1,13 @@
+import { type ComponentPropsWithoutRef, createContext } from "react";
+
+import type { Avatar } from "../avatar";
+
+type AvatarContextType = {
+  size: ComponentPropsWithoutRef<typeof Avatar>["size"];
+};
+
+export const AvatarContext = createContext<{
```
radnan, line 9: let's call this `AvatarGroupContext`
```suggestion
export const AvatarGroupContext = createContext<{
```

packages/react/src/avatar-context/AvatarContext.tsx
```diff
@@ -0,0 +1,13 @@
...
+  size: AvatarContextType["size"] | undefined;
```
radnan, line 10: we can inline the size - no need for an extra `AvatarContextType` type
```suggestion
export const AvatarContext = createContext<{
  size: ComponentPropsWithRef<typeof Avatar>["size"] | undefined;
```

packages/react/src/avatar-group/AvatarGroup.css.ts
```diff
@@ -0,0 +1,9 @@
+import { recipe } from "../vanilla-extract";
+
+export const avatarGroup = recipe({
+  base: [
+    {
+      display: "flex",
```
radnan, line 6: let's move this sprinkle prop inline to the tsx - since there are no complex styles here

packages/react/src/avatar-group/AvatarGroup.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const AvatarGroup = forwardRef<HTMLDivElement, AvatarGroupProps>(
+  ({ children, className, maxItems = 3, size = "md", ...props }, ref) => {
+    const allChildren = Children.toArray(children);
+    const visibleChildren = allChildren.slice(0, maxItems);
+    const overflowChildren = allChildren.slice(maxItems);
+
+    return (
+      <AvatarContext.Provider value={{ size }}>
+        <Box ref={ref} {...styles.avatarGroup({}, className)} {...props}>
+          {visibleChildren}
```
radnan, line 27: instead of wrapping everything with the context provider - let's only wrap the `visibleChildren`
```suggestion
      <Box ref={ref} {...styles.avatarGroup({}, className)} {...props}>
        <AvatarContext.Provider value={{ size }}>
          {visibleChildren}
        </AvatarContext.Provider>
```

packages/react/src/avatar-group/AvatarGroup.css.ts
```diff
@@ -0,0 +1,9 @@
...
+    },
+  ],
+});
```
radnan, line 9: ah i see now why we used this - to share a marker between avatar and avatar-group - let's do this a bit differently:

```suggestion
export const className = "ax-AvatarGroup";

export const avatarGroup = recipe({
  base: [
    className,
    {
      display: "flex",
    },
  ],
});
```

and in `Avatar.css.ts` we can re-use this `className` export

```ts
      selectors: {
        [`${styles.className} &:not(:first-child)`]: {
          marginLeft: "-12px",
        },
      },
```

packages/react/src/avatar/Avatar.css.ts
```diff
@@ -1,6 +1,10 @@
+import { theme } from "../styles";
 import { mapValues } from "../utils";
 import { style } from "../vanilla-extract";
 import { type RecipeVariants, recipe } from "../vanilla-extract";
+import * as styles from "./../avatar-group/AvatarGroup.css";
+
+const group = styles.avatarGroup().className;
```
radnan, line 7: see https://github.com/optimizely-axiom/optiaxiom/pull/222/files#r1718770517

packages/react/src/avatar/Avatar.tsx
```diff
@@ -34,12 +35,15 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(
...
       src,
       ...props
     },
     ref,
   ) => {
+    const avatarSize = useContext(AvatarContext)?.size;
+    const size = _sizeProp || avatarSize || "md";
```
radnan, line 45: let's rename `AvatarContext` to `AvatarGroupContext` and keep the variable as such
```suggestion
    const groupContext = useContext(AvatarGroupContext);
    const size = sizeProp || groupContext?.size || "md";
```

packages/react/src/avatar-group/AvatarGroup.css.ts
toha993: Vai, I need the style for styling in avatar.css

radnan (approved):

PR #224, opened 2024-08-12
Add `DataTable` component

toha993:


packages/react/src/icons/IconSort.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+
+import { withIcon } from "./withIcon";
+
+interface SortIconProps {
+  sortDirection: "asc" | "desc" | false;
```
radnan, line 6: this doesn't look right

we should have 3 separate icons exported - `IconSort`, `IconSortUp`, and `IconSortDown` and conditionally use that in the component rather than have this logic remain within the icon component

packages/react/src/data-table/DataTable.css.ts
```diff
@@ -0,0 +1,61 @@
...
+    {
+      bg: "white",
+    },
+    style({
+      minWidth: columnWidth,
+    }),
+  ],
+  variants: {
+    pinned: {
+      false: style({}),
```
radnan, line 52: let's replace with just `{}` so we don't create blank class names needlessly
```suggestion
      false: {},
```

packages/react/src/data-table/DataTable.css.ts
```diff
@@ -0,0 +1,61 @@
...
+      left: style({
+        left: cellOffset,
+        position: "sticky",
+        zIndex: 1,
+      }),
+      right: style({}),
```
radnan, line 58: same here
```suggestion
      right: {},
```

packages/react/src/data-table/DataTable.css.ts
```diff
@@ -0,0 +1,61 @@
...
+      minWidth: columnWidth,
+    }),
+  ],
+  variants: {
+    pinned: {
+      false: style({}),
+      left: style({
+        left: cellOffset,
+        position: "sticky",
+        zIndex: 1,
```
radnan, line 56: let's use zIndex from sprinkles - so this would be `10`

packages/react/src/data-table/DataTable.css.ts
```diff
@@ -0,0 +1,61 @@
+import { createVar, recipe, style } from "../vanilla-extract";
+
+export const columnWidth = createVar();
+export const cellOffset = createVar();
```
radnan, line 4: vars should have the suffix `Var` to clearly denote they are css vars
```suggestion
export const columnWidthVar = createVar();
export const cellOffsetVar = createVar();
```

packages/react/src/data-table/DataTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { TableHead } from "../table-head";
+import { TableHeaderCell } from "../table-header-cell";
+import { TableRow } from "../table-row";
+import * as styles from "./DataTable.css";
+
+interface DataTableProps<TData, TValue> {
+  columns: ColumnDef<TData, TValue>[];
+  data: TData[];
+  pinnedColumns?: string[];
+}
```
radnan, line 28: this doesn't look right - we should re-use the tanstack Table type - something like:
```suggestion
import type { TableOptions } from "@tanstack/react-table";

type DataTableProps<TData> = TableOptions<TData>;

export const DataTable = <TData,>({ columns, data, ...props }: DataTableProps<TData>) => {
```

packages/react/src/data-table/DataTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+export type { ColumnDef };
```
radnan, line 30: we don't need to re-export this from here - consumers should grab them directly from `@tanstack/react-table`

packages/react/src/data-table/DataTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+            </TableRow>
+          ))}
+        </TableBody>
+      </Table>
+
+      {data.length > pageSize && (
+        <Box mt="8">
+          <Pagination
+            onPageChange={(newPage) => table.setPageIndex(newPage - 1)}
+            page={pageIndex + 1}
```
radnan, line 141: re-use table state -
```suggestion
            page={table.getState().pagination.pageIndex}
```

packages/react/src/data-table/DataTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    {flexRender(cell.column.columnDef.cell, cell.getContext())}
+                  </TableCell>
+                );
+              })}
+            </TableRow>
+          ))}
+        </TableBody>
+      </Table>
+
+      {data.length > pageSize && (
```
radnan, line 137: let's re-use table state
```suggestion
      {data.length > table.getState().pagination.pageSize && (
```

radnan: the reason being the pagination state can be controlled or uncontrolled - and we need to write code that is agnostic of that

packages/react/src/icons/IconSort.tsx
toha993: Gotcha vai, will update it.

packages/react/src/data-table/DataTable.tsx
toha993: Do we need to install `@tanstack/react-table` for storybook then?

radnan (approved):

PR #225, opened 2024-08-12
add  `Layout` component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/187

### DIFF
--- a/.changeset/good-rocks-tap.md
+++ b/.changeset/good-rocks-tap.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/react": minor
+---
+
+add Layout Component
--- a/apps/storybook/src/components/Layout.stories.tsx
+++ b/apps/storybook/src/components/Layout.stories.tsx
@@ -1,0 +1,146 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import {
+  Box,
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+  Heading,
+  Nav,
+  NavAccountItem,
+  NavBody,
+  NavFooter,
+  NavItem,
+  NavList,
+  Sidebar,
+  SidebarToggle,
+  Text,
+} from "@optiaxiom/react";
+import { Layout } from "@optiaxiom/react/unstable";
+import {
+  IconBinaryTree,
+  IconChartInfographic,
+  IconChartLine,
+  IconExternalLink,
+  IconFlag2,
+  IconHistory,
+  IconLayoutSidebar,
+  IconSettings,
+  IconUsers,
+  IconVocabulary,
+} from "@tabler/icons-react";
+
+export default {
+  component: Layout,
+  decorators: (Story) => (
+    <Box style={{ height: "100vh", width: "90vw" }}>
+      <Story />
+    </Box>
+  ),
+} satisfies Meta<typeof Layout>;
+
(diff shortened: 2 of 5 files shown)


packages/react/src/app-shell/AppShell.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const AppShell = forwardRef<HTMLDivElement, AppShellProps>(
+  ({ children, className, header, navbar, ...props }, ref) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
+    return (
+      <Flex
+        ref={ref}
+        {...styles.appShell({}, className)}
+        {...sprinkleProps}
+        {...restProps}
```
toha993, line 25: We can just spread all the props.

packages/react/src/app-shell/AppShell.tsx
```diff
@@ -0,0 +1,42 @@
...
+      >
+        {header}
+        <Flex
+          alignItems="start"
+          flexDirection="row"
+          gap="0"
+          justifyContent="start"
+        >
+          <Flex>{navbar}</Flex>
```
toha993, line 34: Do we need Flex here, I think Box is the most suitable.

packages/react/src/app-shell/AppShell.css.ts
```diff
@@ -0,0 +1,13 @@
+import { recipe, style } from "../vanilla-extract";
+
+export const appShell = recipe({
+  base: [
+    {
+      gap: "0",
+      h: "full",
+    },
+    style({
+      width: "100vh",
```
toha993, line 10: Can't we use `w="full"` for this? or other way the story should handle the width.

packages/react/src/app-shell/AppShell.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Box, type BoxProps } from "../box";
+import { Flex } from "../flex";
+import { extractSprinkles } from "../sprinkles";
+import * as styles from "./AppShell.css";
+
+type AppShellProps = BoxProps<
+  "div",
+  {
+    header?: ReactNode;
+    navbar?: ReactNode;
```
toha993, line 12: let's rename it to SideNav prop

packages/react/src/app-shell/AppShell.tsx
jaf107: I think, we can just skip `Box` or `Flex` and just leave {sidenav} to as it is

toha993: That's more preferred I believe if we are not appending any style.

.changeset/gentle-dots-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/react": minor
+---
+
+Adding new Appshell component
```
toha993, line 5: Squash this commit as changelog links to the commit where the changeset is added - rather than the PR

packages/react/src/app-shell/AppShell.css.ts
```diff
@@ -0,0 +1,10 @@
+import { recipe } from "../vanilla-extract";
+
+export const appShell = recipe({
+  base: [
+    {
+      gap: "0",
+      h: "full",
+    },
+  ],
```
toha993, line 9: Let's inline the sprinkles if we are not doing any complex styling or need of any selector.

radnan (approved):

PR #228, opened 2024-08-12
add `GlobalNav` component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/197


packages/react/src/global-nav/GlobalNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+                role="navigation"
+                {...restProps}
+              >
+                <Flex alignItems="center" gap="4" px="xs" w="full">
+                  <Flex flex="1" gap="4" justifyContent="start" w="full">
+                    {children}
+                  </Flex>
+
+                  <RadixCollapsible.Trigger asChild>
+                    <GlobalNavItem startDecorator={<IconAngleLeft />}>
```
toha993, line 42: Lets use the given icon for collapse instead of IconAngleLeft.

packages/react/src/global-nav/GlobalNav.css.ts
```diff
@@ -0,0 +1,19 @@
+import { recipe } from "../vanilla-extract";
```
toha993, line 1: As there's no complex selectors or styling, we can inline the styling.

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+  organization,
+}: GlobalNavProfileMenuProps) => {
+  return (
+    <Flex {...styles.wrapper({})}>
+      <Box asChild {...styles.picture()}>
+        {avatar}
+      </Box>
+
+      <RadixCollapsible.Content asChild>
+        <Flex flexDirection="row">
```
toha993, line 33:
```suggestion
   <Flex asChild flexDirection="row">
      <RadixCollapsible.Content>
```

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const GlobalNavProfileMenu = ({
+  avatar,
+  name,
+  organization,
+}: GlobalNavProfileMenuProps) => {
+  return (
+    <Flex {...styles.wrapper({})}>
+      <Box asChild {...styles.picture()}>
+        {avatar}
+      </Box>
```
toha993, line 30: Can we reuse the avatar component here?

packages/react/src/global-nav-item/GlobalNavItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { extractSprinkles } from "../sprinkles";
+import * as styles from "./GlobalNavItem.css";
+
+export type GlobalNavItemProps = BoxProps<
+  "button",
+  {
+    endDecorator?: ReactNode;
+    startDecorator?: ReactNode;
+  }
+>;
```
toha993, line 17:
```suggestion
export type GlobalNavItemProps = BoxProps<typeof Button>;
```

packages/react/src/global-nav/GlobalNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { type ReactNode } from "react";
+
+import { Box, type BoxProps } from "../box";
+import { Flex } from "../flex";
+import { GlobalNavContext } from "../global-nav-context";
+import { GlobalNavItem } from "../global-nav-item";
+import { IconAngleLeft } from "../icons/IconAngleLeft";
+import { extractSprinkles } from "../sprinkles";
+import * as styles from "./GlobalNav.css";
+type GlobalNavProps = BoxProps<
```
toha993, line 12: Let's add a Blank line before it.
```suggestion

type GlobalNavProps = BoxProps<
```

packages/react/src/global-nav/GlobalNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+  "nav",
+  {
+    endDecorator?: ReactNode;
+  }
+>;
+
+export const GlobalNav = forwardRef<HTMLElement, GlobalNavProps>(
```
toha993, line 19: Do we also need to make the collapsible controlled from outside vai?

c.c @radnan vai.

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { IconEllipsis } from "../icons/IconEllipsis";
+import { Text } from "../text";
+import * as styles from "./GlobalNavProfileMenu.css";
+
+export type GlobalNavProfileMenuProps = BoxProps<
+  "div",
+  {
+    avatar: ReactNode;
+    name?: string;
+    organization?: string;
```
toha993, line 17: What if we handle them through children like the way we handled label? So, we just wrap the children with user-Info styling, and they can have their own style or related information within the children?

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.css.ts
```diff
@@ -0,0 +1,25 @@
+import { recipe } from "../vanilla-extract";
```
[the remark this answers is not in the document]
jaf107: got it!

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
jaf107: I am afraid we can't. The borderRadius doesn't match here

packages/react/src/global-nav/GlobalNav.css.ts
jaf107: Won't we need to pass className for the 'nav' ?

toha993: It will be automatically handled when we are spreading the props for nav.

packages/react/src/global-nav-item/GlobalNavItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+          active,
+        })}
+        {...sprinkleProps}
+      >
+        <Button
+          appearance="secondary"
+          icon={startDecorator}
+          ref={ref}
+          {...restProps}
+        >
```
toha993, line 37: We can remove the css.ts and have inline styling as button has an `active` state.

```suggestion
      <Flex
        alignItems="center"
        asChild
        flexDirection="row"
        gap="sm"
        h="lg"
        px="sm"
        py="xs"
        w="full"
        {...sprinkleProps}
      >
        <Button
          appearance="secondary"
          data-state={active ? "active" : undefined}
          icon={startDecorator}
          ref={ref}
          {...restProps}
        >
```

packages/react/src/global-nav/GlobalNav.tsx
radnan: we can allow both controlled/uncontrolled by doing something like exposing the following props:

```ts
{
  defaultExpanded?: boolean;
  expanded?: boolean;
  onExpandedChange?: (expanded: boolean) => void;
}
```

packages/react/src/global-nav-item/GlobalNavItem.tsx
jaf107: This was interesting. Thanks!

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+>;
+
+export const GlobalNavProfileMenu = ({
+  avatar,
+  name,
+  organization,
+}: GlobalNavProfileMenuProps) => {
+  return (
+    <Flex alignItems="center" flexDirection="row" gap="xs" px="md" py="xs">
+      <Box asChild rounded="sm" size="md">
```
toha993, line 27: We need to use avatar here and use a `sm` radius for it

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+>;
+
+export const GlobalNavProfileMenu = ({
+  avatar,
+  name,
+  organization,
+}: GlobalNavProfileMenuProps) => {
+  return (
+    <Flex alignItems="center" flexDirection="row" gap="xs" px="md" py="xs">
```
toha993, line 26: For the `gap` the image isn't positioned in the center. We need to set gap={0} and add a `ml` to the Flex with name and organization

packages/react/src/global-nav-item/GlobalNavItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
+    const { expanded } = useContext(GlobalNavContext);
+
+    return (
+      <Flex
+        alignItems="center"
+        asChild
+        flexDirection="row"
+        gap="sm"
```
toha993, line 28: It has no impact as there's only one child with button

packages/react/src/global-nav-item/GlobalNavItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+        h="lg"
+        px="sm"
+        py="xs"
+        w="full"
+        {...sprinkleProps}
+      >
+        <Button
+          appearance="secondary"
```
toha993, line 36: We might need to add a gap here as there's a gap of `12` in figma between icon and text.

packages/react/src/global-nav/GlobalNav.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    return (
+      <GlobalNavContext.Provider value={{ expanded }}>
+        <RadixCollapsible.Root onOpenChange={setExpanded} open={expanded}>
+          <Flex border="1" pb="md" pt="lg">
+            <Box asChild display="flex" {...sprinkleProps}>
+              <nav
+                aria-label="Global Navigation"
+                ref={ref}
+                role="navigation"
```
toha993, line 59: It will cause redundancy for screen readers. 

```
Screen readers will announce the type of role the landmark is. Because of this, you do not need to describe what the landmark is in its label. For example, a declaration of role="navigation" with an of aria-label="Primary navigation" may be announced redundantly as, "primary navigation navigation".
```

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
jaf107: thanks!

packages/react/src/global-nav-profile-menu/GlobalNavProfileMenu.tsx
jaf107: If we're passing `Avatar` as a Reactnode, I think handling the rounded part should be done there. 
Or we can just pass the img src and handle it by wrapping it with the `Avatar` component here.

C.c - @radnan bhai, what do you think?

radnan: i think this is one exception to our general rule here - so we'll use Avatar and explicitly pass `rounded` to get the shape we want

maybe in the future we can add this as a variant to the Avatar component - but it's not needed right now

packages/react/src/global-nav-item/GlobalNavItem.tsx
jaf107: seems to be working

apps/storybook/src/navigation/GlobalNav.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+const Template: Story["render"] = (args) => (
+  <GlobalNav
+    endDecorator={
+      <GlobalNavAccount
+        avatar={
+          <Avatar src="https://images.unsplash.com/photo-1511485977113-f34c92461ad9?ixlib=rb-1.2.1&w=128&h=128&dpr=2&q=80" />
+        }
+      >
+        <Text fontWeight="500">Rhaenyra Targaryen</Text>
+        <Text fontSize="sm">Optimizely</Text>
```
radnan, line 35: let's move name and organization back to props
```suggestion
      <GlobalNavAccount
        avatar={
          <Avatar src="https://images.unsplash.com/photo-1511485977113-f34c92461ad9?ixlib=rb-1.2.1&w=128&h=128&dpr=2&q=80" />
        }
        name="Rhaenyra Targaryen"
        organization="Optimizely"
```

radnan (approved):

PR #229, opened 2024-08-13
Add `Drawer` component

toha993:


packages/react/src/drawer/Drawer.css.ts
```diff
@@ -0,0 +1,71 @@
...
+      z: "popover",
+    },
+    style({
+      position: "fixed",
+    }),
+  ],
+  variants: {
+    position: {
+      bottom: style({
+        borderTop: `1px solid ${theme.colors["border.secondary"]}`,
```
radnan, line 20: let's use `borderT` sprinkle prop for this and the other right/left borders in the other variants

packages/react/src/drawer/Drawer.css.ts
```diff
@@ -0,0 +1,71 @@
...
+        left: 0,
+        right: 0,
+      }),
+      left: style({
+        borderRight: `1px solid ${theme.colors["border.secondary"]}`,
+        bottom: 0,
+        height: "100%",
+        left: 0,
+        top: 0,
+        width: "25%",
```
radnan, line 31: let's use sprinkle props for width here:

```ts
{
  maxW: ["full", "lg"],
  w: ["full", "3/4"],
}
```

and set it to 75% with a max width of 512px - and make it 100% on mobile

packages/react/src/drawer-title/DrawerTitle.tsx
```diff
@@ -0,0 +1,37 @@
...
+>;
+
+export const DrawerTitle = forwardRef<HTMLDivElement, DrawerTitleProps>(
+  ({ children, description, ...props }, ref) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
+    return (
+      <Flex gap="xs" pb="md" pt="lg" px="lg" {...sprinkleProps}>
+        <Heading asChild level="4">
+          <RadixDrawer.Title ref={ref} {...restProps}>
```
radnan, line 24: adding `ref` and `restProps` to the `Title` doesn't feel right - let's add the ref and other props to the outer component instead:
```suggestion
      <Flex gap="xs" pb="md" pt="lg" px="lg" ref={ref} {...props}>
        <Heading asChild level="4">
          <RadixDrawer.Title>
```

radnan: let's do the same thing on `DialogTitle` as well

packages/react/src/drawer-footer/DrawerFooter.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { Flex } from "../flex";
+import * as styles from "./DrawerFooter.css";
+
+type DrawerFooterProps = ComponentPropsWithRef<typeof Flex>;
+
+export const DrawerFooter = forwardRef<HTMLDivElement, DrawerFooterProps>(
+  ({ children, className, ...props }, ref) => {
+    return (
+      <Flex ref={ref} {...styles.footer({}, className)} {...props}>
```
radnan, line 11: let's inline the `footer` styles since they're all sprinkle props

packages/react/src/drawer/Drawer.css.ts
```diff
@@ -0,0 +1,71 @@
...
+      bottom: style({
+        borderTop: `1px solid ${theme.colors["border.secondary"]}`,
+        bottom: 0,
+        left: 0,
+        right: 0,
+      }),
+      left: style({
+        borderRight: `1px solid ${theme.colors["border.secondary"]}`,
+        bottom: 0,
+        height: "100%",
```
radnan, line 28: let's use sprinkle prop of `h: "full"`

radnan (approved):

PR #231, opened 2024-08-13
Add `HoverCard` component

toha993:


packages/react/src/hover-card-content/HoverCardContent.tsx
```diff
@@ -0,0 +1,78 @@
...
+                        width={8}
+                      >
+                        <polygon points="0,0 30,0 15,10" />
+                        <polyline
+                          points="0,0 15,10 30,0"
+                          strokeLinecap="round"
+                          strokeLinejoin="round"
+                          strokeWidth="2.5"
+                        />
+                      </svg>
```
radnan, line 65: not needed in this PR - but todo: we should extract this out to a re-usable component - like `<IconArrow />` or something

radnan (approved):

PR #239, opened 2024-08-19
Add `mode=multiple` in `Combobox` component

toha993:


packages/react/src/command-checkbox-item/CommandCheckboxItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (mode === "multiple") {
+      if (Array.isArray(contextValue)) {
+        setValue(
+          newChecked
+            ? [...contextValue, value]
+            : contextValue.filter((v) => v !== value),
+        );
+      } else {
+        setValue([value]);
+      }
```
radnan, line 37: i don't think we need this isArray check here? we can simplify this
```suggestion
      setValue(
        newChecked
          ? [...contextValue, value]
          : contextValue.filter((v) => v !== value),
      );
```

radnan (approved):

PR #241, opened 2024-08-22
Styling fix for components

toha993:


packages/react/src/input-base/InputBase.css.ts
```diff
@@ -82,11 +78,6 @@ export const input = recipe({
         },
       },
     }),
-    responsiveStyle({
```
radnan, line 85: let's revert this change please - we can add a comment here on why this was done - but the idea here is to explicitly use `16px` on mobile because otherwise iOS will zoom in on the page anytime the input is focused

https://defensivecss.dev/tip/input-zoom-safari/

packages/react/src/input-base/InputBase.css.ts
```diff
@@ -63,10 +58,11 @@ export const input = recipe({
     {
       bg: "transparent",
       flex: "auto",
+      fontWeight: "400",
     },
     style({
-      fontSize: "16px",
-      lineHeight: "22px",
```
radnan, line 69: we need to revert this change - and we can add a comment explaining it - line-height was set to `22px` to adjust for the `1px` border on top and bottom

packages/react/src/input-base/InputBase.css.ts
toha993: Got it vai.

radnan:
this PR seems to have a lot of different fixes all rolled together - can we please split them up at least by component? otherwise it's a bit hard to review the changes

PR #242, opened 2024-08-22
fix `Accordion` as per design

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/180

### DIFF
--- a/packages/react/src/accordion-content/AccordionContent.tsx
+++ b/packages/react/src/accordion-content/AccordionContent.tsx
@@ -13,7 +13,7 @@ export const AccordionContent = forwardRef<
   return (
     <Box asChild ref={ref} {...styles.content({}, className)} {...props}>
       <RadixAccordion.Content>
-        <Box color="fg.default" fontSize="md">
+        <Box color="fg.default" fontSize="md" p="xs" pt="0">
           {children}
         </Box>
       </RadixAccordion.Content>
--- a/packages/react/src/accordion-content/AccordionContent.tsx
+++ b/packages/react/src/accordion-content/AccordionContent.tsx
@@ -13,7 +13,7 @@ export const AccordionContent = forwardRef<
   return (
     <Box asChild ref={ref} {...styles.content({}, className)} {...props}>
       <RadixAccordion.Content>
-        <Box color="fg.default" fontSize="md" pb="md">
+        <Box color="fg.default" fontSize="md">
           {children}
         </Box>
       </RadixAccordion.Content>
--- a/packages/react/src/accordion-item/AccordionItem.tsx
+++ b/packages/react/src/accordion-item/AccordionItem.tsx
@@ -11,7 +11,7 @@ export const AccordionItem = forwardRef<HTMLDivElement, AccordionItemProps>(
     const { restProps, sprinkleProps } = extractSprinkles(props);
 
     return (
-      <Box asChild gap="4" ref={ref} {...sprinkleProps}>
+      <Box asChild ref={ref} {...sprinkleProps}>
         <RadixAccordion.Item value={value} {...restProps}>
           {children}
         </RadixAccordion.Item>
--- a/packages/react/src/accordion-item/AccordionItem.tsx
+++ b/packages/react/src/accordion-item/AccordionItem.tsx
@@ -1,16 +1,22 @@
 import * as RadixAccordion from "@radix-ui/react-accordion";
 import { forwardRef } from "react";
 
-import { Box, type BoxProps } from "../box";
+import { type BoxProps } from "../box";
+import { Flex } from "../flex";
+import { extractSprinkles } from "../sprinkles";
 
 type AccordionItemProps = BoxProps<typeof RadixAccordion.Item>;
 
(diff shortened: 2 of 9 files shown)


packages/react/src/accordion/Accordion.tsx
```diff
@@ -15,7 +15,7 @@ export const Accordion = forwardRef<HTMLDivElement, AccordionProps>(
     const { restProps, sprinkleProps } = extractSprinkles(props);
 
     return (
-      <Box asChild borderB="1" ref={ref} {...sprinkleProps}>
+      <Box asChild ref={ref} w="320" {...sprinkleProps}>
```
toha993, line 18: I think width should be passed from the stories if needed; if not passed it would take the whole container.
```suggestion
      <Box asChild ref={ref} {...sprinkleProps}>
```

packages/react/src/accordion-item/AccordionItem.tsx
```diff
@@ -1,16 +1,22 @@
...
 
 export const AccordionItem = forwardRef<HTMLDivElement, AccordionItemProps>(
   ({ children, value, ...props }, ref) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
     return (
-      <Box asChild borderT="1" ref={ref} {...props}>
-        <RadixAccordion.Item value={value}>{children}</RadixAccordion.Item>
-      </Box>
+      <Flex asChild gap="4" ref={ref} {...sprinkleProps}>
```
radnan, line 15: we cannot use Flex for the accordion item because accordions can collapse and expand. and the way this toggling works is using JS we animate the height of accordion content from 0 to it's intrinsic height when opening and then back to 0 when closing.

but if you use a flexbox and define a gap between the trigger and the content - then you'll notice "jumps" when the content is displayed with a 0 height because at that point flexbox will introduce the gap in between the trigger and the content.

you can see this clearly when you remove text from the content and then toggle it (i've also removed padding from the content to illustrate the jump further)

https://github.com/user-attachments/assets/51453b3b-a7c3-450e-8533-3923e88a164f

packages/react/src/accordion-content/AccordionContent.tsx
```diff
@@ -13,7 +13,7 @@ export const AccordionContent = forwardRef<
   return (
     <Box asChild ref={ref} {...styles.content({}, className)} {...props}>
       <RadixAccordion.Content>
-        <Box color="fg.default" fontSize="md" pb="md">
+        <Box color="fg.default" fontSize="md">
```
radnan, line 16: we have to apply padding to this inner box instead of the outer content because opening/closing the item applies height animation to this outer component - and the padding will throw off the animation - similar to the other flexbox/gap change - it'll introduce a "jump" to the animation.

you can clearly see the jumping in this recording when you remove text from the content:

https://github.com/user-attachments/assets/3ac2be35-a6fa-4779-b78a-c56cf1d8df1f

packages/react/src/accordion-trigger/AccordionTrigger.tsx
```diff
@@ -4,43 +4,51 @@ import { type ReactNode, forwardRef } from "react";
...
 import { IconAngleDown } from "../icons/IconAngleDown";
+import { IconAngleRight } from "../icons/IconAngleRight";
 import { extractSprinkles } from "../sprinkles";
 import * as styles from "./AccordionTrigger.css";
 
 type AccordionTriggerProps = BoxProps<
   typeof RadixAccordion.Trigger,
   {
     endDecorator?: ReactNode;
+    position?: "end" | "start";
```
radnan, line 15: having both start/end decorations and a `position` prop does not make sense here. we also should not allow toggling the position at the individual trigger level - rather this should be an appearance or variant prop at the very top `<Accordion />` component level

for now let's change this to an `appearance` prop and call it `primary` (chevron at the start) and `secondary` (chevron at the end)

we can use context to pass this appearance prop down to the triggers and then inside the trigger if a `startDecorator` or `endDecorator` is explicitly given we'll use that, else we'll fallback to `IconAngleDown` at the appropriate position.

packages/react/src/accordion-item/AccordionItem.tsx
```diff
@@ -2,14 +2,19 @@ import * as RadixAccordion from "@radix-ui/react-accordion";
...
 type AccordionItemProps = BoxProps<typeof RadixAccordion.Item>;
 
 export const AccordionItem = forwardRef<HTMLDivElement, AccordionItemProps>(
   ({ children, value, ...props }, ref) => {
+    const { restProps, sprinkleProps } = extractSprinkles(props);
+
     return (
-      <Box asChild borderT="1" ref={ref} {...props}>
-        <RadixAccordion.Item value={value}>{children}</RadixAccordion.Item>
+      <Box asChild gap="4" ref={ref} {...sprinkleProps}>
```
radnan, line 14: remove unused gap
```suggestion
      <Box asChild ref={ref} {...sprinkleProps}>
```

radnan (approved):

toha993:
Keeping it for future design instruction. 

https://github.com/radix-ui/primitives/discussions/583

PR #249, opened 2024-08-25
fix `Button` as per figma

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/34

### DIFF
--- a/apps/storybook/src/forms/Button.stories.tsx
+++ b/apps/storybook/src/forms/Button.stories.tsx
@@ -13,7 +13,11 @@ export default {
 
 type Story = StoryObj<typeof Button>;
 
-const sizes = ["sm", "md", "lg"] as const;
+const sizes = [
+  ["sm", "Small"],
+  ["md", "Medium"],
+  ["lg", "Large"],
+] as const;
 const appearances = [
   "default",
   "primary",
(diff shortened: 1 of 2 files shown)


apps/storybook/src/forms/Button.stories.tsx
```diff
@@ -13,92 +13,104 @@ export default {
...
-  ...Variants,
+export const Disabled: Story = {
   args: {
-    colorScheme: "danger",
+    children: "Button",
+    disabled: true,
   },
+  render: (args) => (
+    <Flex flexDirection="row">
+      {sizes.map((size) => (
```
radnan, line 64: can we showcase the disabled state of all the different appearances instead of the sizes here?
```suggestion
      {appearances.map((appearance) => (
```

apps/storybook/src/forms/Button.stories.tsx
```diff
@@ -13,92 +13,104 @@ export default {
...
+export const Sizes: Story = {
   args: {
-    colorScheme: "primary",
+    children: "Button",
   },
+  render: (args) => (
+    <Flex flexDirection="row">
+      {sizes.map((size) => (
+        <Button {...args} key={size} size={size}>
+          {size}
```
radnan, line 50: let's write out the full size label here (small, medium, large) instead of using the shorthand prop value
```
const sizes = [
  ["sm", "Small"],
  ["md", "Medium"],
  ["lg", "Large"],
] as const;
```
```suggestion
      {sizes.map(([size, label]) => (
        <Button {...args} key={size} size={size}>
          {label}
```

apps/storybook/src/forms/Button.stories.tsx
```diff
@@ -13,92 +13,104 @@ export default {
...
+        <Button {...args} key={size} size={size}>
+          Button
+        </Button>
+      ))}
+    </Flex>
+  ),
 };
 
-export const Upload: Story = {
+export const StandaloneIcon: Story = {
```
radnan, line 73: let's rename this story to something like `IconOnly`
```suggestion
export const IconOnly: Story = {
```

apps/storybook/src/forms/Button.stories.tsx
```diff
@@ -13,92 +13,104 @@ export default {
...
+    <Flex flexDirection="row">
+      {sizes.map((size) => (
+        <Button {...args} key={size} size={size} />
+      ))}
     </Flex>
   ),
 };
 
-export const appearance: Story = {
+export const Icons: Story = {
```
radnan, line 86: and let's rename this story to something like `IconsWithText`
```suggestion
export const IconsWithText: Story = {
```

apps/storybook/src/forms/Button.stories.tsx
jaf107: Got it bhai!

radnan (approved):

PR #254, opened 2024-08-27
Update styling form `Avatar`

toha993:


packages/react/src/avatar/Avatar.css.ts
```diff
@@ -44,16 +44,16 @@ export const avatar = recipe({
...
       }),
     ),
     size: {
       xs: { fontSize: "xs", size: "xs" },
-      sm: { fontSize: "sm", size: "sm" },
+      sm: { fontSize: "xs", size: "sm" },
       md: { fontSize: "md", size: "md" },
-      lg: { fontSize: "lg", size: "lg" },
-      xl: { fontSize: "xl", size: "xl" },
+      lg: { fontSize: "2xl", size: "48" },
```
radnan, line 55: let's change the existing `xl` size token from 64 to 48 and use it here
```suggestion
      lg: { fontSize: "2xl", size: "xl" },
```

packages/react/src/avatar/Avatar.css.ts
```diff
@@ -44,16 +44,16 @@ export const avatar = recipe({
...
+      xl: { fontSize: "4xl", size: "80" },
```
radnan, line 56: let's add a new `5xl` size token for 80 and use it here
```suggestion
      xl: { fontSize: "4xl", size: "5xl" },
```

packages/react/src/avatar/Avatar.css.ts
```diff
@@ -18,10 +18,10 @@ export const avatar = recipe({
 
       selectors: {
         [`${styles.className} &:not(:first-child)`]: {
-          marginLeft: "-12px",
+          marginLeft: "-3px",
         },
         [`${styles.className} &`]: {
-          border: `2px solid ${theme.colors["white"]}`,
+          outline: `1px solid ${theme.colors["white"]}`,
```
radnan, line 24: let's switch back to border here since we want the element to fit within the given size - outline will spill the content outside the bounding box.

packages/react/src/avatar/Avatar.css.ts
```diff
@@ -44,16 +44,16 @@ export const avatar = recipe({
...
       }),
     ),
     size: {
       xs: { fontSize: "xs", size: "xs" },
-      sm: { fontSize: "sm", size: "sm" },
+      sm: { fontSize: "xs", size: "sm" },
       md: { fontSize: "md", size: "md" },
-      lg: { fontSize: "lg", size: "lg" },
-      xl: { fontSize: "xl", size: "xl" },
+      lg: { fontSize: "2xl", size: "xl" },
```
radnan, line 55:
```suggestion
      lg: { fontSize: "2xl", size: "lg" },
```

packages/react/src/avatar/Avatar.tsx
```diff
@@ -32,7 +30,7 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(
     {
       children,
       className,
-      colorScheme = "neutral",
+      colorScheme = "brand",
```
radnan, line 33: let's keep the default as neutral 
```suggestion
      colorScheme = "neutral",
```

packages/react/src/avatar/Avatar.css.ts
```diff
@@ -69,14 +67,23 @@ export const fallback = recipe({
...
-      xs: { px: "4" },
-      sm: { px: "6" },
-      md: { px: "8" },
-      lg: { px: "10" },
-      xl: { px: "20" },
+      xs: style({ width: "12px" }),
+      sm: style({ width: "14px" }),
+      md: style({ width: "16px" }),
+      xl: style({ width: "20px" }),
+      "5xl": style({ width: "30px" }),
```
radnan, line 84: let's use sprinkle props for these? and shift the sizes slightly to match the existing tokens for now:

```suggestion
      xs: { w: "12" },
      sm: { w: "12" },
      md: { w: "16" },
      xl: { w: "20" },
      "5xl": { w: "32" },
```

radnan (approved):

PR #256, opened 2024-08-27
Update styling for `Dialog` title

toha993:


packages/react/src/dialog-title/DialogTitle.tsx
```diff
@@ -21,13 +21,13 @@ export const DialogTitle = forwardRef<HTMLDivElement, DialogTitleProps>(
 
     return (
       <Flex gap="xs" pb="md" pt="lg" px="lg" {...sprinkleProps}>
-        <Heading asChild level="4">
+        <Heading asChild fontSize="3xl" fontWeight="500" level="4">
```
radnan, line 24: if this is a level 3 heading then let's use that - so for `Heading` we should avoid modifying the `fontSize` directly as much as possible and rely on the `level`/`variant` prop instead
```suggestion
        <Heading asChild fontWeight="500" level="3">
```

packages/react/src/alert-dialog/AlertDialog.css.ts
```diff
@@ -63,8 +63,8 @@ export const footer = recipe({
 export const title = recipe({
   base: [
     {
-      fontSize: "2xl",
-      fontWeight: "600",
+      fontSize: "3xl",
+      fontWeight: "500",
```
radnan, line 67: can we re-use the `Heading` component instead of setting the `fontSize` directly?

so in the `AlertDialog.tsx` file we should have something like:

```tsx
<Heading asChild fontWeight="600" level="3" p="24" pb="16">
  <RadixAlertDialog.Title>{title}</RadixAlertDialog.Title>
</Heading>
```

radnan (approved):

PR #258, opened 2024-08-27
Update styling for `InputBase`

toha993:


packages/react/src/input-base/InputBase.css.ts
```diff
@@ -14,27 +14,23 @@ export const wrapper = recipe({
       border: "1",
       flexDirection: "row",
       gap: "0",
-      rounded: "sm",
+      rounded: "md",
     },
     style({
       borderColor: theme.colors["border.default"],
-      color: theme.colors["fg.default"],
+      color: theme.colors["bg.neutral.inverse"],
```
radnan, line 21: let's update the `fg.default` token itself to point to `neutral.800` rather than using a background color token for the text color here

so in `colors.ts` let's do this change:

```diff
diff --git a/packages/react/src/tokens/colors.ts b/packages/react/src/tokens/colors.ts
index 93d55c01..87c9c15f 100644
--- a/packages/react/src/tokens/colors.ts
+++ b/packages/react/src/tokens/colors.ts
@@ -176,7 +176,7 @@ export const colors = {
   /**
    * Primary text color for page headings
    */
-  "fg.default": colorPalette["neutral.900"],
+  "fg.default": colorPalette["neutral.800"],
   /**
    * Primary text color when used on solid color bg
    */
```

packages/react/src/input-base/InputBase.css.ts
```diff
@@ -63,8 +59,12 @@ export const input = recipe({
     {
       bg: "transparent",
       flex: "auto",
+      fontWeight: "400",
```
radnan, line 62: 400 is the default font-weight - don't think this is going to do anything, right?

radnan (approved):

PR #267, opened 2024-08-30
Add `pill` in combobox

toha993:


toha993:
Won't support it for now.

PR #276, opened 2024-09-04
Move `badge` to feedback stories

toha993:


apps/storybook/src/feedback/Badge.stories.tsx
radnan, line 1: let's also update it in docs

https://github.com/optimizely-axiom/optiaxiom/blob/515a4bddf278b0fafe1269f52a8f86741aebc1c2/apps/docs/pages/components/_meta.json#L48C4-L48C9

move it to the bottom "Feedback" section after "Alert"

radnan (approved):

Issue #265, opened 2024-08-29
remove colorScheme and variant from Button props


radnan:
only use it inside the css recipe

PR #277, opened 2024-09-04
remove `colorScheme` and `variant` from Button props

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/265

### DIFF
--- a/packages/react/src/toast-action/ToastAction.tsx
+++ b/packages/react/src/toast-action/ToastAction.tsx
@@ -18,7 +18,7 @@ export const ToastAction = forwardRef<HTMLButtonElement, ToastProps>(
           orientation="vertical"
         />
 
-        <Button appearance="secondary" asChild size="sm" {...props}>
+        <Button appearance="inverse" asChild size="sm" {...props}>
           <RadixToast.Action altText={altText} ref={ref}>
             {children}
           </RadixToast.Action>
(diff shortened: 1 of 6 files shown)


packages/react/src/toast-action/ToastAction.tsx
```diff
@@ -18,7 +18,7 @@ export const ToastAction = forwardRef<HTMLButtonElement, ToastProps>(
           orientation="vertical"
         />
 
-        <Button asChild size="sm" variant="solid" {...props}>
+        <Button appearance="secondary" asChild size="sm" {...props}>
```
radnan, line 21: let's add a new appearance for these called `inverse`

so `inverse` will use `colorScheme="neutral"` and `variant="solid"`

packages/react/src/toast/Toast.tsx
```diff
@@ -46,10 +46,10 @@ export const Toast = forwardRef<HTMLLIElement, ToastProps>(
 
           <RadixToast.Close asChild>
             <Button
+              appearance="secondary"
```
radnan, line 49: same comment - use the new `inverse` appearance here

packages/react/src/toast-action/ToastAction.tsx
jaf107: should we show it in storybook bhai?

radnan: Nah we don’t need to show it there - the toast stories will cover them

radnan (approved):

PR #295, opened 2024-09-06
chore: update pnpm and all packages

radnan:


package.json
```diff
@@ -14,27 +14,27 @@
...
     "esbuild": "0.17.6",
     "happy-dom": "^14.12.3",
     "prettier": "^3.3.3",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
-    "rollup": "^4.19.0",
+    "rollup": "^4.21.2",
     "rollup-plugin-dts": "^6.1.1",
     "rollup-plugin-esbuild": "^6.1.1",
     "typescript": "^5.5.4",
```
radnan, line 36: explicitly holding back `typescript` update to `5.6` as it conflicts with our existing version of `@typescript-eslint/typescript-eslint` plugin (which we still have to use because we're stuck using eslint v8 because other plugins don't yet work with eslint v9)

package.json
```diff
@@ -14,27 +14,27 @@
...
   "devDependencies": {
-    "@changesets/cli": "^2.27.7",
+    "@changesets/cli": "^2.27.8",
     "@optiaxiom/shared": "workspace:^",
     "@rollup/pluginutils": "^5.1.0",
-    "@testing-library/jest-dom": "^6.4.8",
+    "@testing-library/jest-dom": "^6.5.0",
     "@testing-library/react": "^14.3.1",
     "@testing-library/user-event": "^14.5.2",
     "@types/react": "^18.3.3",
```
radnan, line 23: explicitly holding back `@types/react` update to `18.3.5` as it conflicts with type generation for `Transition` component props fro some reason (guessing it has to do with the `data-size` prop)

PR #299, opened 2024-09-08
Remove `sm` size and add `backgroundColor` for `InputBase`

toha993:


packages/react/src/input-base/InputBase.css.ts
```diff
@@ -17,6 +17,7 @@ export const wrapper = recipe({
       rounded: "md",
     },
     style({
+      backgroundColor: theme.colors["white"],
```
radnan, line 20: Let's add this using sprinkles? And use `surface` instead of white to support light/dark mode

toha993: Vai, we are changing backgroundColor in selectors for `disabled`, that's why I added it in style.

radnan: Ah didn’t notice that - let’s just change it to surface here then
```suggestion
      backgroundColor: theme.colors["surface"],
```

radnan (approved):

PR #300, opened 2024-09-08
Update stories and fix `gap` based on orientation for `RadioGroup`

toha993:


packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -13,12 +13,26 @@ const mapDirectionToOrientation = {
...
   ({ children, flexDirection = "column", ...props }, ref) => {
     const { restProps, sprinkleProps } = extractSprinkles(props);
 
     return (
-      <Flex asChild flexDirection={flexDirection} gap="sm" {...sprinkleProps}>
+      <Flex
+        asChild
+        flexDirection={flexDirection}
+        gap={
+          flexDirection === "row" || flexDirection === "column"
```
radnan, line 30: We have to use `mapResponsiveValue` here instead of just a simple map

Example:

https://github.com/optimizely-axiom/optiaxiom/blob/1a85ead9590fbe71af505616c0b1192cf0665ef3/packages/react/src/flex/Flex.tsx#L37-L40

radnan (approved):

PR #301, opened 2024-09-08
Fix endDecorator color for `ControlBase`

toha993:


packages/react/src/control-base/ControlBase.tsx
```diff
@@ -29,7 +29,6 @@ export const ControlBase = forwardRef<HTMLDivElement, ControlBaseProps>(
   ) => {
     const { restProps, sprinkleProps } = extractSprinkles(props);
     const id = useId(idProp);
-
```
radnan, line 32: Please keep the bank line here

We should always use blank lines to group related blocks of code - kinda like paragraphs in text

packages/react/src/control-base/ControlBase.tsx
```diff
@@ -44,7 +43,11 @@ export const ControlBase = forwardRef<HTMLDivElement, ControlBaseProps>(
         </Text>
 
         {endDecorator && (
-          <Box asChild {...styles.decorator()}>
+          <Box
+            asChild
+            color={props.disabled ? "fg.disabled" : "fg.secondary"}
```
radnan, line 48: Let's add the fontSize here as well

In majority cases I'm assuming we'll need small text

packages/react/src/control-base/ControlBase.tsx
toha993: My fault vai, I added a log in this line, later removed the log and it happened.

apps/storybook/src/forms/Checkbox.stories.tsx
```diff
@@ -15,11 +15,7 @@ export const Basic: Story = {};
 
 export const HelperText: Story = {
   args: {
-    endDecorator: (
-      <Text color="fg.secondary" fontSize="sm">
-        Helper Text
-      </Text>
-    ),
+    endDecorator: <Text>Helper Text</Text>,
```
radnan, line 18: technically we can just change these to strings
```suggestion
    endDecorator: "Helper Text",
```

apps/storybook/src/forms/RadioGroup.stories.tsx
```diff
@@ -125,7 +107,7 @@ export const ComplexExample1: Story = {
         <Flex className={styles.item} {...props} gap="xs">
           <RadioGroupItem
             endDecorator={
-              <Text color="fg.secondary" fontSize="sm">
+              <Text color="fg.secondary">
                 Displays a message as a bar at top of page
               </Text>
```
radnan, line 112: let's change these to strings as well
```suggestion
              "Displays a message as a bar at top of page"
```

packages/react/src/control-base/ControlBase.tsx
radnan: can you bring back the blank line in that case please

radnan: also needs a rebase with `main`

radnan (approved):

PR #315, opened 2024-09-12
Add `Dropdown` Sub-Menu components

toha993:


packages/react/src/unstable.ts
```diff
@@ -18,6 +18,9 @@ export * from "./command-input";
 export * from "./command-item";
 export * from "./command-list";
 export * from "./command-separator";
+export * from "./dropdown-sub";
+export * from "./dropdown-sub-content";
+export * from "./dropdown-sub-trigger";
```
radnan, line 23: these should be called `DropdownMenu` + `Sub`
```suggestion
export * from "./dropdown-menu-sub";
export * from "./dropdown-menu-sub-content";
export * from "./dropdown-menu-sub-trigger";
```

radnan (approved):
need to rename the components - but looks good otherwise

PR #319, opened 2024-09-12
refactor input and add PopoverAnchor component

radnan:


packages/react/src/popover-anchor/PopoverAnchor.tsx
```diff
@@ -0,0 +1,20 @@
+import * as RadixPopover from "@radix-ui/react-popover";
+import { forwardRef } from "react";
+
+import { Box, type BoxProps } from "../box";
+
+type PopoverAnchorProps = BoxProps<typeof RadixPopover.Root>;
```
toha993, line 6: Vai, Will it have the same props of Root or Anchor?

radnan: fixed

PR #353, opened 2024-09-22
Add `TooltipRoot` `TooltipTrigger` `TooltipContent` component

toha993:


packages/react/src/unstable.ts
```diff
@@ -28,3 +28,6 @@ export * from "./table-cell";
 export * from "./table-head";
 export * from "./table-header-cell";
 export * from "./table-row";
+export * from "./tooltip-content";
```
radnan, line 31: we can move this to stable

packages/react/src/tooltip-trigger/TooltipTrigger.tsx
```diff
@@ -0,0 +1,42 @@
...
+type TooltipTriggerProps = ComponentPropsWithoutRef<
+  typeof RadixTooltip.Trigger
+>;
+
+export const TooltipTrigger = forwardRef<
+  HTMLButtonElement,
+  TooltipTriggerProps
+>(({ asChild, children, ...props }, outerRef) => {
+  const { keepOpenOnActivation, setOpen, triggerRef } =
+    useTooltipContext("Tooltip");
```
radnan, line 18:
```suggestion
    useTooltipContext("TooltipTrigger");
```

packages/react/src/tooltip-content/TooltipContent.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const TooltipContent = forwardRef<HTMLDivElement, TooltipContentProps>(
+  (
+    { align = "center", children, className, sideOffset = 2, ...props },
+    outerRef,
+  ) => {
+    const innerRef = useRef<HTMLDivElement>(null);
+    const ref = useComposedRefs(innerRef, outerRef);
+
+    const { keepOpenOnActivation, open } = useTooltipContext("Tooltip");
```
radnan, line 23:
```suggestion
    const { keepOpenOnActivation, open } = useTooltipContext("TooltipContent");
```

packages/react/src/tooltip-context/TooltipContext.ts
```diff
@@ -0,0 +1,10 @@
+import type { RefObject } from "react";
+
+import { createContext } from "@radix-ui/react-context";
+
+export const [TooltipContextProvider, useTooltipContext] = createContext<{
+  keepOpenOnActivation?: boolean;
+  open?: boolean;
+  setOpen: (open: boolean) => void;
+  triggerRef: RefObject<HTMLButtonElement>;
+}>("Tooltip");
```
radnan, line 10:
```suggestion
}>("TooltipRoot");
```

packages/react/src/tooltip-content/TooltipContent.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { theme } from "../theme";
+import { useTooltipContext } from "../tooltip-context";
+import { Transition } from "../transition";
+import * as styles from "./TooltipContent.css";
+
+type TooltipContentProps = BoxProps<typeof RadixTooltip.Content>;
+
+export const TooltipContent = forwardRef<HTMLDivElement, TooltipContentProps>(
+  (
+    { align = "center", children, className, sideOffset = 2, ...props },
```
radnan, line 17: let's move the z=popover to here
```suggestion
    { align = "center", children, className, sideOffset = 2, z = "popover", ...props },
```

packages/react/src/tooltip/Tooltip.tsx
```diff
@@ -44,118 +32,29 @@ export const Tooltip = forwardRef<HTMLButtonElement, TooltipProps>(
       delayDuration,
       keepOpenOnActivation,
       onOpenChange,
-      open: openProp,
-      withArrow,
+      open,
       z = "popover",
```
radnan, line 36: we can remove this - since `TooltipContent` already supports `z` sprinkle prop and uses `z: "popover" in the css.ts file
```suggestion
```

packages/react/src/tooltip-content/TooltipContent.css.ts
```diff
@@ -0,0 +1,27 @@
...
+      bg: "bg.neutral.inverse",
+      color: "white",
+      px: "12",
+      py: "8",
+      rounded: "lg",
+      z: "popover",
+    },
+    style({
+      position: "relative",
+    }),
```
radnan, line 17: why do we need this position=relative ?

packages/react/src/tooltip-content/TooltipContent.tsx
```diff
@@ -0,0 +1,65 @@
...
+                        }
+                      : undefined
+                  }
+                  ref={ref}
+                  sideOffset={sideOffset}
+                >
+                  <Text fontSize="sm">{children}</Text>
+
+                  <RadixTooltip.Arrow
+                    fill={theme.colors["neutral.900"]}
```
radnan, line 51: should be same as tooltip bg
```suggestion
                    fill={theme.colors["bg.neutral.inverse"]}
```

packages/react/src/tooltip/Tooltip.tsx
```diff
@@ -44,118 +32,29 @@ export const Tooltip = forwardRef<HTMLButtonElement, TooltipProps>(
...
-                      />
-                    )}
-                  </RadixTooltip.Content>
-                </Box>
-              </Transition>
-            </RadixTooltip.Portal>
-          )}
-        </AnimatePresence>
-      </RadixTooltip.Root>
+        <TooltipContent sideOffset={5} z={z} {...props}>
```
radnan, line 54: let's move this sideOffset=5 default to the `TooltipContent` component (which currently has sideOffset=2)

and let's remove the `z` from here
```suggestion
        <TooltipContent {...props}>
```

packages/react/src/tooltip-content/TooltipContent.css.ts
toha993: No vai, we can remove it.

packages/react/src/tooltip-content/TooltipContent.css.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const content = recipe({
+  base: [
+    {
+      bg: "bg.neutral.inverse",
+      color: "white",
+      px: "12",
+      py: "8",
+      rounded: "lg",
+      z: "popover",
```
radnan, line 11: let's add a new zIndex for tooltips and use a higher value - like `10000`
```suggestion
      z: "tooltip",
```

and in `packages/globals/src/tokens/zIndex.ts` add an entry for `tooltip: "10000"`

packages/react/src/tooltip-content/TooltipContent.tsx
radnan: this needs to be addressed still

radnan (approved):

packages/react/src/tooltip-content/TooltipContent.tsx
toha993: Missed it vai.

PR #359, opened 2024-09-23
Remove `bg` for sub menu trigger on `data-state=open` state

toha993:


radnan:
let's also move the DropdownMenuSub... components to stable

packages/react/src/index.ts
```diff
@@ -40,6 +40,9 @@ export * from "./dropdown-menu-content";
 export * from "./dropdown-menu-item";
 export * from "./dropdown-menu-label";
 export * from "./dropdown-menu-separator";
+export * from "./dropdown-menu-sub";
```
radnan, line 43: we need to add these to `packages/web-components/package.json` file

radnan (approved):

PR #369, opened 2024-09-25
Add icon in `AlertDialogTitle`

toha993:


jaf107:
Let's handle the changes in one single commit

jaf107:
Let's rebase & get this merged

jaf107 (approved):

packages/react/src/alert-dialog-title/AlertDialogTitle.tsx
```diff
@@ -1,28 +1,36 @@
...
-      level="3"
-      p="lg"
-      pb="md"
-      ref={ref}
-      {...props}
-    >
-      <RadixAlertDialog.Title>{children}</RadixAlertDialog.Title>
-    </Heading>
+    <Flex flexDirection="row" gap="md" p="lg" pb="md" {...sprinkleProps}>
+      <Box asChild color="fg.accent.red" h="16" w="auto">
```
toha993, line 19: Need to pass addonBefore

toha993:
Will merge after #385

.changeset/chilled-lemons-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/react": minor
```
radnan, line 2: should be a patch type change - since we're still on v0 - so while this is a new backwards compatible feature - for now we're considering both bug fixes and features as patch - and breaking changes as minor
```suggestion
"@optiaxiom/react": patch
```

toha993: Got it vai.

.changeset/chilled-lemons-shave.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Allow consumers to have a custom or default icon in alert dialog
```
radnan, line 5: let's re-word this a bit - something like:
```suggestion
added customizable icon to alert dialog title via `addonBefore` prop
```

Issue #370, opened 2024-09-26
add `ToggleGroup` component


radnan:
https://www.radix-ui.com/primitives/docs/components/toggle-group

PR #376, opened 2024-09-29
add `SegmentedControl` Component

jaf107:
resolves: https://github.com/optimizely-axiom/optiaxiom/issues/370


packages/react/src/toggle-group-item/ToggleGroupItem.tsx
```diff
@@ -0,0 +1,29 @@
...
+  ToggleGroupItemProps
+>(({ children, disabled, ...props }, ref) => {
+  const { restProps, sprinkleProps } = extractSprinkles(props);
+
+  return (
+    <Button
+      appearance="secondary"
+      asChild
+      disabled={disabled}
+      {...sprinkleProps}
```
radnan, line 20: let's forward all props here? don't think we need to extract sprinkles
```suggestion
      {...props}
```

packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { forwardRef } from "react";
+
+import { Box, type BoxProps } from "../box";
+
+type ToggleGroupProps = Omit<BoxProps<typeof RadixToggleGroup.Root>, "type">;
+
+export const ToggleGroup = forwardRef<HTMLDivElement, ToggleGroupProps>(
+  ({ children, ...props }, ref) => {
+    return (
+      <Box asChild {...props}>
```
radnan, line 11: i think we should use `ButtonGroup` here - or at the very least a Flex

packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -0,0 +1,20 @@
+import * as RadixToggleGroup from "@radix-ui/react-toggle-group";
+import { forwardRef } from "react";
+
+import { Box, type BoxProps } from "../box";
+
+type ToggleGroupProps = Omit<BoxProps<typeof RadixToggleGroup.Root>, "type">;
```
radnan, line 6: why are we omitting `type` here?

apps/storybook/src/forms/ToggleGroup.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+export const Basic: Story = {
+  render: () => {
+    return (
+      <ToggleGroup>
+        <ToggleGroupItem value="a">A</ToggleGroupItem>
+        <ToggleGroupItem value="b">B</ToggleGroupItem>
+        <ToggleGroupItem value="c">C</ToggleGroupItem>
+      </ToggleGroup>
+    );
+  },
```
radnan, line 23: let's move the children to the `default` args and avoid using `render` in the stories

packages/react/src/toggle-group/ToggleGroup.tsx
jaf107: it supports multiple where we can toggle multiple buttons of the group. 
To me, it seems like we're going for toggling only one button among the group. For this reason, we're omitting `type`

packages/react/src/toggle-group/ToggleGroup.tsx
jaf107: Using `ButtonGroup` here with 0 gap will give a line separating the buttons. Using `Flex` for this bhai

apps/storybook/src/forms/ToggleGroup.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type Story = StoryObj<typeof ToggleGroup>;
+
+export const Basic: Story = {
+  args: {
+    children: (
+      <>
+        <ToggleGroupItem icon={<IconLayoutGrid />} value="a" />
+        <ToggleGroupItem icon={<IconLayoutGrid />} value="b" />
+        <ToggleGroupItem icon={<IconLayoutGrid />} value="c" />
```
radnan, line 21: Screenshot 2024-09-30 at 3 13 42 PM

let's use one of these two as an example here - rather than just 3 identical grid icons

packages/react/src/toggle-group/ToggleGroup.tsx
radnan: let's keep the `type` here - so consumers can decide whether they want single or multiple

we can default it to single from our side - so something like:
```tsx
export function ToggleGroup({ type = "single" }) {
  return <RadixToggleGroup.Root type={type} />
}
```

apps/storybook/src/forms/ToggleGroup.stories.tsx
jaf107: This seems more applicable. Will update it bhai.
Thanks!

radnan (approved):

PR #378, opened 2024-10-01
Add `Autocomplete` component

toha993:


apps/figma/package.json
radnan, line 1: could you drop this commit please?

apps/storybook/src/overlays/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,297 @@
...
+    return (
+      <Autocomplete
+        isItemDisabled={isItemDisabled}
+        itemToKey={itemToKey}
+        itemToString={itemToString}
+        items={items}
+        onInputValueChange={onInputValueChange}
+      >
+        <AutocompleteTrigger>
+          <AutocompleteLabel label="Label">
```
radnan, line 201: we don't need this `AutocompleteLabel` component - we can get rid of it entirely

apps/storybook/src/overlays/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,297 @@
...
+const itemToKey = (book: unknown) => {
+  return (book as Book).id;
+};
+
+const isItemDisabled = (book: unknown) => {
+  return (book as Book).disabled;
+};
+
+export const Basic: Story = {
+  render: function Basic() {
```
radnan, line 141: let's accept args here
```suggestion
  render: function Basic(args) {
```

and forward it to the `Autocomplete` component

then in your `Disabled` story you can just change it to something like:

```ts
export const Disabled: Story = {
  ...Basic,
  args: {
    disabled: true,
  },
};
```

apps/storybook/src/overlays/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,297 @@
...
+        </AutocompleteContent>
+      </Autocomplete>
+    );
+  },
+};
+
+export const DisabledItems: Story = {
+  render: function DefaultSelected() {
+    const [items, setItems] = useState(books);
+    const [value, setValue] = useState("Crime and Punishment");
```
radnan, line 258: let's have a separate story for `Value` - also the value/onValueChange types are not correct - these are supposed to be the same as the `items[number]` type
```suggestion
    const [value, setValue] = useState(books[0]);
```

apps/storybook/src/overlays/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,297 @@
...
+    const onInputValueChange = (inputValue: string) => {
+      setItems(
+        books.filter(
+          (book: Book) =>
+            !inputValue ||
+            book.title.includes(inputValue) ||
+            book.author.includes(inputValue),
+        ),
+      );
+    };
```
radnan, line 272: let's inline these props:

```
<Autocomplete
  onInputValueChange={(inputValue) => setItems(...)}
>
```

apps/storybook/src/overlays/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+export const DisabledItems: Story = {
+  render: function DefaultSelected() {
+    const [items, setItems] = useState(books);
+    const [value, setValue] = useState("Crime and Punishment");
+
+    function onValueChange(value: string) {
+      setValue(value);
+    }
+    const onInputValueChange = (inputValue: string) => {
```
radnan, line 263: let's forward the args as-is in `onInputValueChange`

so this would be
```suggestion
    const onInputValueChange = ({ inputValue }) => {
```

packages/react/src/autocomplete-content/AutocompleteContent.css.ts
```diff
@@ -0,0 +1,39 @@
+import { recipe, style } from "../vanilla-extract";
+
+export const content = recipe({
+  base: [
+    style({
+      maxHeight: "var(--radix-popover-content-available-height)",
+      minWidth: "var(--radix-popover-trigger-width)",
+      position: "relative",
```
radnan, line 8: doesn't popover already have this - do we need to duplicate this here?

packages/react/src/autocomplete-content/AutocompleteContent.css.ts
```diff
@@ -0,0 +1,39 @@
...
+});
+
+export const wrapperList = recipe({
+  base: [
+    style({
+      listStyle: "none",
+      overflow: "auto",
+      overscrollBehavior: "contain",
+      transition: "100ms ease",
+      transitionProperty: "height",
```
radnan, line 36: why do we need this?

packages/react/src/autocomplete-content/AutocompleteContent.css.ts
```diff
@@ -0,0 +1,39 @@
...
+        z: "popover",
+      },
+    },
+  },
+});
+
+export const wrapperList = recipe({
+  base: [
+    style({
+      listStyle: "none",
```
radnan, line 32: isn't this set in Box? don't think we need this here

packages/react/src/autocomplete-content/AutocompleteContent.css.ts
toha993: We aren't reusing our popover rather the radix popover as we need to force-mount.

packages/react/src/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,88 @@
...
+  {
+    disabled?: boolean;
+    initialHighlightedIndex?: number;
+    isItemDisabled?: (item: unknown) => boolean;
+    itemToKey?: (item: unknown) => string;
+    itemToString?: (item: unknown) => string;
+    items: unknown[];
+    onInputValueChange: (inputValue: string) => void;
+    onValueChange?: (value: string) => void;
+    value?: string;
```
radnan, line 18: these should be unknown - generic Item argument
```suggestion
    onValueChange?: (value: Item) => void;
    value?: Item;
```

packages/react/src/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,88 @@
...
+      : value,
+    isItemDisabled(item) {
+      return isItemDisabled ? isItemDisabled(item) : false;
+    },
+    itemToKey(item) {
+      return itemToKey ? itemToKey(item) : item;
+    },
+    itemToString(item) {
+      return itemToString ? itemToString(item) : String(item);
+    },
```
radnan, line 50: let's always pass these as-is without modification
```suggestion
    isItemDisabled,
    itemToKey,
    itemToString,
```

packages/react/src/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,88 @@
...
+        case useCombobox.stateChangeTypes.InputBlur:
+          downshift.selectItem(downshift.selectedItem ?? "");
+          break;
+      }
+    },
+
+    onSelectedItemChange({ selectedItem }) {
+      onValueChange?.(
+        itemToString ? itemToString(selectedItem) : (selectedItem as string),
+      );
```
radnan, line 67: let's pass `selectedItem` as-is
```suggestion
      onValueChange?.(selectedItem);
```

radnan: related to this we need a way to clear the value - but we can do that in another PR - need some input from designers

packages/react/src/autocomplete-content/AutocompleteContent.css.ts
toha993: Added it from CommandItem.css.ts, will remove it vai.

packages/react/src/autocomplete-context/AutocompleteContext.ts
```diff
@@ -0,0 +1,13 @@
+import type { UseComboboxReturnValue } from "downshift";
+
+import { createContext } from "@radix-ui/react-context";
+
+export const [AutocompleteContextProvider, useAutocompleteContext] =
+  createContext<{
+    disabled?: boolean;
+    downshift: UseComboboxReturnValue<unknown>;
+    highlightedItem: unknown;
+    itemToString?: (item: unknown) => string;
```
radnan, line 10: why do we need this in the context?

apps/storybook/src/overlays/Autocomplete.stories.tsx
radnan: actually now that i think about it - the `disabled` prop should go on the `AutocompleteInput` element - not on the `Autocomplete` element

packages/react/src/autocomplete-context/AutocompleteContext.ts
```diff
@@ -0,0 +1,13 @@
+import type { UseComboboxReturnValue } from "downshift";
+
+import { createContext } from "@radix-ui/react-context";
+
+export const [AutocompleteContextProvider, useAutocompleteContext] =
+  createContext<{
+    disabled?: boolean;
```
radnan, line 7: we shouldn't have this here - disabled should be set on the input directly if the intent is to disable the input

packages/react/src/autocomplete-context/AutocompleteContext.ts
```diff
@@ -0,0 +1,13 @@
...
+    downshift: UseComboboxReturnValue<unknown>;
+    highlightedItem: unknown;
+    itemToString?: (item: unknown) => string;
+    items: unknown[];
+    onValueChange?: (value: string) => void;
```
radnan, line 12: we shouldn't need these two here right?

radnan (approved):

PR #388, opened 2024-10-02
make min-width configurable on menu content base

### LINKED ISSUE #381: make minWidth on `<DropdownMenuContent />` configurable instead of hard coded

radnan:
resolves #381

### DIFF
--- a/packages/react/src/menu-content-base/MenuContentBase.tsx
+++ b/packages/react/src/menu-content-base/MenuContentBase.tsx
@@ -16,13 +16,13 @@ export type MenuContentBaseProps<
   ExtendProps<
     {
       open: boolean | undefined;
-    },
+    } & styles.ContentVariants,
     P
   >
 >;
 
 export const MenuContentBase = forwardRef<HTMLDivElement, MenuContentBaseProps>(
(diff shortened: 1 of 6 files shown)


packages/react/src/menu-content-base/MenuContentBase.tsx
```diff
@@ -1,20 +1,28 @@
...
+      open: boolean | undefined;
+    } & styles.ContentVariants,
+    P
+  >
 >;
 
-export const MenuContentBase = forwardRef<HTMLDivElement, MenuContentProps>(
-  ({ children, className, open, ...props }, ref) => {
+export const MenuContentBase = forwardRef<HTMLDivElement, MenuContentBaseProps>(
+  ({ children, className, minW, open, ...props }, ref) => {
```
toha993, line 25: Don't we need to set a default value for it vai?

radnan: no - by default it'll be ignored and it won't set any `minWidth` style at all

but if you want you can set it to `trigger` to match the trigger width - or once you've already done that you can reset it back by overriding it with `0`

PR #400, opened 2024-10-06
Add `aria-label` in `SearchInput` and update `aria-label` for `Sidenav`

toha993:


packages/react/src/search-input/SearchInput.tsx
```diff
@@ -30,6 +30,7 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchProps>(
           value && (
             <Button
               appearance="subtle"
+              aria-label="clear"
```
radnan, line 33: let's use capital case here
```suggestion
              aria-label="Clear"
```

also i see that this button isn't focusable in native search box of browser - so we should do the same

let's add a `tabIndex={-1}` to this button

packages/react/src/side-nav/SideNav.tsx
```diff
@@ -64,7 +64,7 @@ export const SideNav = forwardRef<HTMLElement, SideNavProps>(
             transition={animations ? "all" : undefined}
             w={expanded ? "224" : "56"}
           >
-            <nav aria-label="Global Navigation" ref={ref}>
+            <nav aria-label="side" ref={ref}>
```
radnan, line 67: let's do Main Navigation here
```suggestion
            <nav aria-label="Main Navigation" ref={ref}>
```

just `side` by itself won't be very meaningful

another alternative could be `Side navigation`

toha993: Vai, the navigation role would trigger the screen reader side + navigation 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#best_practices

radnan (approved):

PR #401, opened 2024-10-06
Add zIndex for `Button` on focus-visible

toha993:


packages/react/src/button/Button.css.ts
```diff
@@ -41,6 +41,7 @@ export const button = recipe({
           {
             outline: `2px solid ${theme.colors["outline.brand"]}`,
             outlineOffset: "1px",
+            zIndex: "1",
```
radnan, line 44: let's use a theme token here
```suggestion
            zIndex: themes.zIndex["10"],
```

radnan (approved):

PR #402, opened 2024-10-06
Add justified prop in `Tabs`

toha993:


packages/react/src/tabs/Tabs.tsx
```diff
@@ -8,15 +8,15 @@ import * as styles from "./Tabs.css";
 
 type TabsProps = BoxProps<
   typeof RadixTabs.Root,
-  { appearance?: "primary" | "secondary" }
+  { appearance?: "primary" | "secondary"; justified?: boolean }
```
radnan, line 11: let's do multiple lines here
```suggestion
  {
    appearance?: "primary" | "secondary";
    justified?: boolean;
  }
```

radnan: also was justified the name we finalized on? i thought it was something like `equalWidth`

IMO this should be automatic if possible based on the current screen size

so if possible let's do it in CSS - otherwise do it based on media query - but let's not expose this prop to the consumers

apps/storybook/src/navigation/Tabs.stories.tsx
```diff
@@ -52,6 +52,9 @@ export default {
     w: "384",
   },
   component: Tabs,
+  parameters: {
+    layout: "padded",
+  },
```
radnan, line 57: don't think we need this change published - this was just for testing locally
```suggestion
```

radnan (approved):

PR #404, opened 2024-10-07
Add downshift `Select` component

toha993:


packages/react/src/select-value/SelectValue.tsx
```diff
@@ -0,0 +1,21 @@
...
+type SelectValueProps = BoxProps<"div", { placeholder?: string }>;
+
+export const SelectValue = forwardRef<HTMLDivElement, SelectValueProps>(
+  ({ children, placeholder, ...props }, ref) => {
+    const { downshift, itemToString } = useSelectContext("SelectValue");
+    return (
+      <Box {...props} ref={ref}>
+        {(children ?? downshift.selectedItem)
+          ? itemToString(downshift.selectedItem)
+          : placeholder}
```
radnan, line 15: this doesn't look right - it should be something like:
```suggestion
        {children ??
          (downshift.selectedItem
            ? itemToString(downshift.selectedItem)
            : placeholder)}
```

packages/react/src/select/Select.tsx
```diff
@@ -0,0 +1,84 @@
...
+  items,
+  itemToKey = (value) => value,
+  itemToString = (value) => (value ? String(value) : ""),
+  onValueChange,
+  value,
+  ...props
+}: SelectProps<Item>) {
+  const { id: toggleButtonId } = useFieldContext();
+
+  const [selectedItem, setSelectedItem] = useControllableState({
```
radnan, line 38: since we're now controlling the `selectedItem` state we need to pass it down in the context instead of relying on `downshift.selectedItem` which won't have the up-to-date value always

packages/react/src/select-value/SelectValue.tsx
```diff
@@ -0,0 +1,21 @@
+import { forwardRef } from "react";
+
+import { Box, type BoxProps } from "../box";
+import { useSelectContext } from "../select-context";
+
+type SelectValueProps = BoxProps<"div", { placeholder?: string }>;
+
+export const SelectValue = forwardRef<HTMLDivElement, SelectValueProps>(
+  ({ children, placeholder, ...props }, ref) => {
+    const { downshift, itemToString } = useSelectContext("SelectValue");
```
radnan, line 10: following up on https://github.com/optimizely-axiom/optiaxiom/pull/404/files#r1803088222

our own `selectedItem` should be passed here rather than picked from downshift

```suggestion
    const { downshift, itemToString, selectedItem } = useSelectContext("SelectValue");
```

packages/react/src/select/Select.tsx
radnan: it'll be late by 1 render - so on the next render you'll have the correct value - but in the render where the state changes you'll have a stale value

radnan (approved):

PR #405, opened 2024-10-07
Restructure `Autocomplete` component

toha993:


packages/react/src/autocomplete-empty/AutocompleteEmpty.tsx
```diff
@@ -2,24 +2,25 @@ import { type ComponentPropsWithRef, forwardRef } from "react";
...
-      </Box>
-    );
-  },
-);
+export const AutocompleteEmpty = forwardRef<
+  HTMLDivElement,
+  AutocompleteEmptyProps
+>(({ children, ...props }, ref) => {
+  return (
+    <Box
```
radnan, line 12: this should be an `li` same as `AutocompleteItem`

packages/react/src/autocomplete-empty/AutocompleteEmpty.tsx
```diff
@@ -2,24 +2,25 @@ import { type ComponentPropsWithRef, forwardRef } from "react";
...
-        p="md"
-        ref={ref}
-        {...props}
-      >
-        {children || "No results found"}
-      </Box>
-    );
-  },
-);
+export const AutocompleteEmpty = forwardRef<
```
radnan, line 7: i'm wondering if we should call this `AutocompleteEmptyItem`

so under this umbrella you have `Item` and `EmptyItem`

packages/react/src/autocomplete-item/AutocompletItem.tsx
```diff
@@ -8,47 +8,48 @@ import { extractSprinkles } from "../sprinkles";
...
-          {children}
+        {children}
 
-          {downshift.selectedItem === item && <IconCheck />}
-        </li>
-      </Flex>
-    );
-  },
-);
+        {downshift.selectedItem === item && <IconCheck />}
```
radnan, line 49: can we make this optional? maybe move this to an `ItemIndicator` component

so `AutocompleteItemIndicator` - and consumers can use them if they want or opt-out

and all it does is render an icon if the parent item is the currently selected item

so we'd need an additional context here to pass like `active` or something

```
<AutocompleteItemContextProvider active={downshift.selectedItem === item}>
  <li>{...}</li>
</AutocompleteItemContextProvider>
```

and in `AutocompleteItemIndicator` we'd render null or icon depending on this context value

packages/react/src/autocomplete-item/AutocompletItem.tsx
```diff
@@ -8,47 +8,48 @@ import { extractSprinkles } from "../sprinkles";
...
-          {addonBefore && (
-            <Box asChild h="16" w="auto">
-              {fallbackSpan(addonBefore)}
-            </Box>
-          )}
+  return (
+    <Flex asChild {...styles.item()} {...sprinkleProps}>
+      <li
+        data-disabled={
+          downshift.getItemProps({ item })["aria-disabled"] ? "" : undefined
```
radnan, line 31: let's extract the itemProps as a separate variable
```
const itemProps = downshift.getItemProps({ item })

return (
  <li data-disabled={itemProps['aria-disabled']}  {...itemProps} />
)
```

radnan (approved):

PR #409, opened 2024-10-08
Restructure `Autocomplete` component

toha993:
- Merge AutocompleteInput with AutocompleteTrigger
- Add Interactive Tests
- Add async story
- Add loading prop in AutocompleteContent
- Add onValueClear prop in `SearchInput`


packages/react/src/search-input/SearchInput.tsx
```diff
@@ -38,12 +49,13 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchProps>(
...
                 forceValueChange(innerRef.current, "");
                 innerRef.current.focus();
+                onValueClear?.();
               }}
               {...styles.clear()}
             />
           )
         }
-        addonBefore={<IconMagnifyingGlass />}
+        addonBefore={withSearchIcon && <IconMagnifyingGlass />}
```
radnan, line 59: i don't think we should provide a dedicated prop to clear the icon - consumers shouldn't be doing this 

and internally we can just send `addonBefore={null}` to override it ourselves

packages/react/src/autocomplete-input/AutocompleteInput.tsx
```diff
@@ -19,12 +19,12 @@ export const AutocompleteInput = forwardRef<
...
     <Input
       addonAfter={
-        downshift.inputValue &&
+        downshift.selectedItem &&
         !disabled && (
           <Button
             appearance="subtle"
             aria-label="clear"
-            icon={downshift.inputValue && <IconX />}
+            icon={downshift.selectedItem && <IconX />}
```
radnan: we can drop this conditional - since we already evaluate this when adding the addon itself
```suggestion
            icon={<IconX />}
```

toha993: Squashed the commit vai, it was committed before.

radnan: hmm - in that case we might want to switch back to a regular input - cause we don't want the cross to show up unless a value has been selected - whereas the regular search input will show cross whenever it has value

radnan: or we could use the `{...(selected && { addonAfter: null })}` trick i suppose

radnan (approved):

PR #421, opened 2024-10-11
add storybook test runner

radnan:


apps/storybook/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "storybook build --docs",
     "dev": "wait-on ../../packages/react/dist/index.d.ts ../../packages/react/dist/unstable.d.ts && storybook dev -p 6006",
-    "lint": "oas-lint"
+    "lint": "oas-lint",
+    "test": "vitest"
```
radnan, line 10: Skip storybook tests on `main`

radnan, line 10: Fix aria-label and key issues in stories

PR #428, opened 2024-10-15
Update styling for `Dialog`

toha993:


toha993:
Move `AlertDialog` and `Dialog` maxHeight to content.

packages/react/src/dialog-content/DialogContent.css.ts
```diff
@@ -4,12 +4,15 @@ export const content = recipe({
...
       bg: "bg.default",
+      display: "flex",
+      flexDirection: "column",
       rounded: "lg",
       shadow: "md",
       z: "popover",
     },
     style({
       left: "50%",
+      maxHeight: "50dvh",
```
radnan, line 15: let's bump up the height here
```suggestion
      maxHeight: "75dvh",
```

we can leave AlertDialog at 50dvh - since it has a slightly different use case

radnan (approved):

PR #441, opened 2024-10-17
re-use icon component in button to fix the icon size

radnan:
icon height should be fixed at 16px and the width will be automatically set according to the icon's aspect ratio

this will allow us to use regular square icons or our font awesome 20:16 icons

### DIFF
--- a/packages/react/src/alert-dialog-title/AlertDialogTitle.tsx
+++ b/packages/react/src/alert-dialog-title/AlertDialogTitle.tsx
@@ -1,7 +1,7 @@
 import * as RadixAlertDialog from "@radix-ui/react-alert-dialog";
 import { forwardRef, type ReactNode } from "react";
 
-import { Box, type BoxProps } from "../box";
+import { type BoxProps } from "../box";
 import { Flex } from "../flex";
 import { Heading } from "../heading";
 import { Icon } from "../icon";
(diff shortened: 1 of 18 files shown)


packages/react/src/alert-dialog-title/AlertDialogTitle.tsx
```diff
@@ -23,12 +23,9 @@ export const AlertDialogTitle = forwardRef<
   const { restProps, sprinkleProps } = extractSprinkles(props);
   return (
     <Flex flexDirection="row" gap="md" p="lg" pb="md" {...sprinkleProps}>
-      {addonBefore && <Icon>{fallbackSpan(addonBefore)}</Icon>}
-      {!addonBefore && (
-        <Box asChild color="fg.error" h="16" w="auto">
-          <IconTriangleExclamation />
-        </Box>
-      )}
+      <Icon asChild color="fg.error">
```
toha993, line 26: About the custom icon, would we always have the same colored? As CMP's icos are not in `red` color.

radnan: it's fine to set it as a default - clients can re-color/re-size them as they require

toha993: Got it, we updated Icon component with `asChild`, before it wasn't.

PR #443, opened 2024-10-17
Change props type for `Field` component

toha993:


packages/react/src/field/Field.tsx
```diff
@@ -82,7 +82,7 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
                   />
                 </HoverCardTrigger>
                 <HoverCardContent side="top" sideOffset={5}>
-                  {info}
+                  {fallbackSpan(info)}
```
radnan, line 85: same thing here - don't need fallbackSpan since the parent `HoverCardContent` element does not have `asChild` prop set

radnan (approved):

PR #446, opened 2024-10-20
Update Angle Icons

toha993:


packages/react/src/icons/IconAngleLeft.tsx
```diff
@@ -5,12 +5,14 @@ export const IconAngleLeft = withIcon(
     fill: "none",
     height: 16,
     name: "IconAngleLeft",
+    viewBox: "0 0 20 16",
     width: 20,
```
radnan, line 9: 20/16 is the default in withIcon - so we can leave them out

```suggestion
    name: "IconAngleLeft",
```

radnan (approved):

PR #449, opened 2024-10-20
Update `Field` icon and add `error` icon for `Field`

toha993:


packages/react/src/field/Field.tsx
```diff
@@ -77,10 +77,11 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
                   <Button
                     aria-label="Information hover"
                     border="0"
-                    h="12"
+                    color="fg.secondary"
+                    h="16"
                     icon={<IconCircleQuestion />}
                     p="0"
-                    w="12"
+                    w="20"
```
radnan, line 84: icon buttons need to be square - let's change this to 16 and use `size` (remove the extra `h`) so it's clear that the intent is for this to be a square
```suggestion
                    size="16"
```

radnan (approved):

packages/react/src/field/Field.tsx
toha993: Okay vai

packages/react/src/field/Field.tsx
```diff
@@ -103,9 +106,12 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
         </FieldContext.Provider>
 
         {error && (
-          <Box asChild color="fg.error" fontSize="sm">
+          <Flex color="fg.error" flexDirection="row" fontSize="sm" gap="4">
+            <Box asChild {...styles.icon()}>
```
radnan, line 110: let's do align start on the flex and use margin-top on the icon to align it vertically - that way we can handle multi-line error messages

also we need a flex none on the icon - so it doesn't shrink when there's a lot of text

```suggestion
          <Flex alignItems="start" color="fg.error" flexDirection="row" fontSize="sm" gap="4">
            <Box asChild flex="none" mt="2" {...styles.icon()}>
```

Screenshot 2024-10-21 at 4 58 30 PM

packages/react/src/field/Field.css.ts
```diff
@@ -0,0 +1,8 @@
+import { recipe, style } from "../vanilla-extract";
+
+export const icon = recipe({
+  base: style({
+    height: "12px",
+    width: "14px",
```
radnan, line 6: we don't need a css.ts file here - cause you can just do `h="12" w="auto"` to achieve the same thing in sprinkles

packages/react/src/field/Field.tsx
```diff
@@ -103,9 +106,12 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
...
+              <IconTriangleExclamationSmall />
+            </Box>
             {fallbackSpan(error)}
```
radnan, line 113: technically we no longer need to wrap the error message in a span - since we're no longer using box composition here
```suggestion
            {error}
```

packages/react/src/icons/IconTriangleExclamationSmall.tsx
```diff
@@ -0,0 +1,16 @@
+import { withIcon } from "./withIcon";
+
+export const IconTriangleExclamationSmall = withIcon(
+  {
+    height: 12,
+    name: "IconTriangleExclamationSmall",
```
radnan, line 6: is this small icon actually needed? i can barely tell the difference between this and the regular `IconTriangleExclamation` icon

Screenshot 2024-10-21 at 5 04 19 PM

toha993: I also thought the same, not sure why designer wanted different svg vai. Should I avoid it vai?

packages/react/src/field/Field.tsx
toha993: Okay vai

PR #454, opened 2024-10-22
Add `Chevron` in autocomplete trigger

toha993:


packages/react/src/autocomplete-trigger/AutocompleteTrigger.tsx
```diff
@@ -37,3 +79,11 @@ export const AutocompleteTrigger = forwardRef<
 });
 
 AutocompleteTrigger.displayName = "@optiaxiom/react/AutocompleteTrigger";
+
+function forceValueChange(input: HTMLInputElement, value: string) {
```
radnan, line 83: damn - didn't realize we'd need to bring in so much code - now i'm thinking if we should keep re-using SearchInput and just modify SearchInput to support additional `addonAfter` (so more things in addition to its own cross button)

toha993: I think we can do the update in searchInput vai.

packages/react/src/autocomplete-trigger/AutocompleteTrigger.tsx
```diff
@@ -2,32 +2,61 @@ import { PopperAnchor } from "@radix-ui/react-popper";
...
+import { Button } from "../button";
+import { Flex } from "../flex";
+import { IconAngleDown } from "../icons/IconAngleDown";
+import { IconX } from "../icons/IconX";
+import { Input } from "../input";
 import { extractSprinkles } from "../sprinkles";
 
-type AutocompleteTriggerProps = ComponentPropsWithRef<typeof SearchInput>;
+type AutocompleteTriggerProps = {
+  onValueClear?: () => void;
```
radnan, line 13: i don't think we need to support this onValueClear prop? autocomplete already has `value` and `onValueChange` props

.changeset/wicked-sheep-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/react": patch
+---
+
+Allow consumers to trigger `Autocomplete` with toggle button
```
radnan, line 5: let's re-word this a bit - maybe say "added toggle button to autocomplete trigger" or something

PR #463, opened 2024-10-24
Add support for Orientaion prop  in `RadioGroup` component

jaf107:


packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -21,13 +21,16 @@ const mapGapToOrientation = {
...
+    { children, flexDirection = "column", orientation = "vertical", ...props },
+    ref,
+  ) => {
     const { restProps, sprinkleProps } = extractSprinkles(props);
 
     return (
       <Flex
         asChild
-        flexDirection={flexDirection}
+        flexDirection={orientation === "vertical" ? "column" : "row"}
```
radnan, line 33: we need to also forward the `orientation` prop to `RadixRadio.RadioGroup`

packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -21,25 +16,24 @@ const mapGapToOrientation = {
...
         gap={mapResponsiveValue(
           flexDirection,
```
radnan, line 30: we also need to address orientation here:
```suggestion
          flexDirection ?? orientation === "vertical" ? "column" : "row",
```

jaf107: addressed bhai

radnan (approved):

PR #467, opened 2024-10-25
replace radix checkbox/radio/switch with native input

radnan:


packages/react/src/radio-group-item/RadioGroupItem.tsx
```diff
@@ -1,37 +1,92 @@
...
-  typeof RadixRadio.Item,
+  typeof CheckboxInputHiddenInput,
   {
     description?: ReactNode;
   }
 >;
 
-export const RadioGroupItem = forwardRef<HTMLDivElement, RadioGroupItemProps>(
-  ({ children, description, id, value, ...props }, ref) => {
+export const RadioGroupItem = forwardRef<HTMLInputElement, RadioGroupItemProps>(
```
radnan, line 21: todo: rename this to just `Radio` since we can now also use this independently of `RadioGroup`

packages/react/src/checkbox-input-hidden-input/CheckboxInputHiddenInput.tsx
```diff
@@ -0,0 +1,38 @@
+import { VisuallyHidden } from "@radix-ui/react-visually-hidden";
+import clsx from "clsx";
+import { type ComponentPropsWithoutRef, forwardRef } from "react";
+
+import type { ExtendProps } from "../utils";
+
+import * as styles from "./CheckboxInputHiddenInput.css";
+
+type CheckboxInputHiddenInputProps = ExtendProps<
+  ComponentPropsWithoutRef<"input">,
```
radnan, line 10: need a good way to document the props/fact that these are built on native input

PR #470, opened 2024-10-29
Fix accessibility

toha993:
- Breadcrumbs
- Spinner
- WithIcon

### DIFF
--- a/packages/react/src/icons/withIcon.tsx
+++ b/packages/react/src/icons/withIcon.tsx
@@ -26,7 +26,6 @@ export const withIcon = (
         fill={fill}
         height={height}
         ref={ref}
-        role="img"
         viewBox={viewBox ?? `0 0 ${width} ${height}`}
         width={width}
         xmlns="http://www.w3.org/2000/svg"
--- a/packages/react/src/spinner/Spinner.tsx
+++ b/packages/react/src/spinner/Spinner.tsx
@@ -30,13 +30,13 @@ export const Spinner = forwardRef<HTMLDivElement, SpinnerProps>(
           aria-label="Loading"
           display="inline-flex"
           ref={ref}
+          role="status"
           {...props}
         >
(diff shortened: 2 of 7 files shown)


packages/react/src/spinner/Spinner.tsx
```diff
@@ -25,15 +25,10 @@ export const Spinner = forwardRef<HTMLDivElement, SpinnerProps>(
...
-          ref={ref}
-          {...props}
-        >
+        <Flex alignItems="center" display="inline-flex" ref={ref} {...props}>
           <Box animation="spin" asChild size={size}>
             <svg
+              aria-label="Loading"
               fill="none"
               height={16}
               role="img"
```
radnan, line 34: let's remove the `img` role here

packages/react/src/spinner/Spinner.tsx
```diff
@@ -25,15 +25,10 @@ export const Spinner = forwardRef<HTMLDivElement, SpinnerProps>(
...
-          alignItems="center"
-          aria-label="Loading"
-          display="inline-flex"
-          ref={ref}
-          {...props}
-        >
+        <Flex alignItems="center" display="inline-flex" ref={ref} {...props}>
           <Box animation="spin" asChild size={size}>
             <svg
+              aria-label="Loading"
```
radnan, line 31: let's move the label back on the flex - and add `role="status"` on the root flex element

packages/react/src/icons/withIcon.tsx
```diff
@@ -23,6 +23,7 @@ export const withIcon = (
   const Icon = forwardRef<SVGSVGElement, ComponentPropsWithoutRef<"svg">>(
     (props, ref) => (
       <svg
+        aria-hidden
```
radnan, line 26: similar comment here - let's just remove the `role="img"` instead of adding an `aria-hidden`

radnan (approved):

PR #474, opened 2024-10-29
improve a11y for alert/banner components

radnan:


packages/react/src/banner/Banner.tsx
```diff
@@ -31,28 +33,38 @@ export const Banner = forwardRef<HTMLDivElement, BannerProps>(
...
+
     return (
-      <Flex ref={ref} {...styles.banner({ colorScheme }, className)} {...props}>
-        <Icon asChild {...styles.icon({ colorScheme })}>
-          {createElement(iconMap[colorScheme])}
-        </Icon>
+      <BannerContextProvider descriptionId={descriptionId} labelId={labelId}>
+        <Flex
+          ref={ref}
+          role="alert"
```
toha993, line 43: Is alert the correct role for it vai? or we can use `role="banner"` here?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role

radnan: no - banner is basically a replacement for the `<header>` element - and is more suited for a global header/topbar or `Skip to content` link

toha993: Got it vai.

packages/react/src/banner/Banner.tsx
```diff
@@ -31,28 +33,38 @@ export const Banner = forwardRef<HTMLDivElement, BannerProps>(
...
+    const labelId = useId();
+
     return (
-      <Flex ref={ref} {...styles.banner({ colorScheme }, className)} {...props}>
-        <Icon asChild {...styles.icon({ colorScheme })}>
-          {createElement(iconMap[colorScheme])}
-        </Icon>
+      <BannerContextProvider descriptionId={descriptionId} labelId={labelId}>
+        <Flex
+          ref={ref}
```
toha993, line 42: Don't we need this for banner also vai? 

```
 aria-describedby={descriptionId}
 aria-labelledby={labelId}
```

radnan: oops - added

PR #479, opened 2024-10-30
Update `Pagination` component

toha993:


packages/react/src/pagination/Pagination.tsx
```diff
@@ -76,14 +67,13 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
...
+                      py="8"
                       size="md"
                     >
-                      <Box alignItems="end" display="flex">
-                        <Box asChild>
-                          <IconEllipsis />
-                        </Box>
-                      </Box>
+                      <Icon asChild>
+                        <IconEllipsis />
```
radnan, line 75: Use `icon` prop instead of children for icon only buttons?

packages/react/src/pagination/Pagination.tsx
```diff
@@ -76,14 +67,13 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
                       alignItems="stretch"
                       display="flex"
                       justifyContent="center"
-                      py="10"
+                      px="6"
```
radnan, line 70: And remove these padding overrides once we use an icon only button

packages/react/src/pagination/Pagination.tsx
toha993: Vai, did you mean using `Button` component instead of `Icon` for the ellipsis?

radnan: ow my bad - i thought this was a button - nvm

toha993:
Add `disabled` prop in `Tooltip`

packages/react/src/tooltip/Tooltip.tsx
```diff
@@ -52,7 +53,9 @@ export const Tooltip = forwardRef<HTMLButtonElement, TooltipProps>(
           {children}
         </TooltipTrigger>
 
-        {!empty && <TooltipContent {...props}>{content}</TooltipContent>}
+        {!disabled && !empty && (
```
radnan, line 56: need to add `disabled` to the `disableHoverableContent` prop as well
```
        disableHoverableContent={disabled || empty}
```

packages/react/src/tooltip/Tooltip.tsx
```diff
@@ -16,6 +16,7 @@ type TooltipProps = BoxProps<
     content?: ReactNode;
     defaultOpen?: boolean;
     delayDuration?: ComponentPropsWithRef<typeof TooltipRoot>["delayDuration"];
+    disabled?: boolean;
```
radnan, line 19: please add a separate changeset with this commit about how we've added the disabled prop to tooltip

PR #500, opened 2024-11-03
Add `fullscreen` variant in `Dialog` component

toha993:


packages/react/src/dialog-content/DialogContent.css.ts
```diff
@@ -22,14 +19,27 @@ export const content = recipe({
...
+        maxWidth: "calc(100dvw - 2 * 24px)",
         width: "600px",
       }),
       lg: style({
+        borderRadius: "8px",
+        maxHeight: "75dvh",
+        maxWidth: "calc(100dvw - 2 * 24px)",
         width: "800px",
       }),
+      full: style({
```
radnan, line 39: let's call it `fullscreen` instead of just `full` and add a story for it

packages/react/src/dialog-content/DialogContent.css.ts
```diff
@@ -22,14 +19,27 @@ export const content = recipe({
...
+        height: "100%",
+        width: "100%",
```
radnan, line 41: let's be explicit and use `100dvw` and `100dvh` here

packages/react/src/dialog-content/DialogContent.css.ts
```diff
@@ -22,14 +19,27 @@ export const content = recipe({
   variants: {
     size: {
       sm: style({
+        borderRadius: "8px",
+        maxHeight: "75dvh",
+        maxWidth: "calc(100dvw - 2 * 24px)",
```
radnan, line 24: let's move this back to the base style instead of repeating it

and reset the styles in the fullscreen branch - like use `0` for radius and `none` for max-height/max-width

radnan (approved):

PR #504, opened 2024-11-04
Spread all props to Radix Trigger

toha993:
https://github.com/radix-ui/primitives/issues/3203


radnan (approved):
need to add a changeset

PR #517, opened 2024-11-07
add Controlled story for `Disclosure` component

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/515

### DIFF
--- a/apps/storybook/src/data-display/Disclosure.stories.tsx
+++ b/apps/storybook/src/data-display/Disclosure.stories.tsx
@@ -6,6 +6,7 @@ import {
   DisclosureTrigger,
   Flex,
 } from "@optiaxiom/react";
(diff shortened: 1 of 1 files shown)


apps/storybook/src/data-display/Disclosure.stories.tsx
```diff
@@ -48,3 +49,17 @@ export const ChevronPosition: Story = {
     </Flex>
   ),
 };
+
+export const Controlled: Story = {
+  render: function DefaultSelected(args) {
```
radnan, line 54: rename this to a generic name like Story or Sample or something related to this story like Controlled
```suggestion
  render: function Controlled(args) {
```

radnan (approved):

PR #518, opened 2024-11-07
Fix `Combobox` styling

toha993:


packages/react/src/combobox-item/ComboboxItem.tsx
```diff
@@ -39,6 +39,7 @@ export const ComboboxItem = forwardRef<HTMLLIElement, ComboboxItemProps>(
                   Array.isArray(value) ? value.includes(item) : value === item
                 }
                 disabled={itemProps["aria-disabled"]}
+                gap="0"
```
radnan, line 42: handled this in #525

packages/react/src/combobox-input/ComboboxInput.tsx
```diff
@@ -14,6 +15,7 @@ export const ComboboxInput = forwardRef<HTMLInputElement, ComboboxInputProps>(
 
     return (
       <Input
+        addonBefore={<IconMagnifyingGlass />}
```
radnan, line 18: Need to move this to the CommandInput

packages/react/src/listbox-item-base/ListboxItemBase.css.ts
```diff
@@ -16,7 +17,8 @@ export const item = recipe({
       flexDirection: "row",
       fontSize: "md",
       gap: "xs",
-      p: "xs",
+      px: "xs",
+      py: "10",
```
radnan, line 21: should we use `minHeight: theme.size.lg` instead here? what is the menu item supposed to look like with 2 lines of text?

radnan: addressed in #537 with `line-height`

radnan (approved):

PR #519, opened 2024-11-07
Add `SelectGroup` component

toha993:


apps/storybook/src/comboboxes/Select.stories.tsx
jaf107, line 1: Can we keep the input data & story components separately?

toha993: we will address it in different PR, as we are using more or less same data in `Select`, `Autocomplete`, `Combobox` components.

packages/react/src/select-group-context/SelectGroupContext.ts
```diff
@@ -0,0 +1,6 @@
+import { createContext } from "@radix-ui/react-context";
+
+export const [SelectGroupContextProvider, useSelectGroupContext] =
+  createContext<{
+    id?: string;
+  }>("SelectGroup");
```
radnan, line 6: we need to provide a default value here otherwise SelectLabel cannot be used outside of SelectGroup
```suggestion
  }>("SelectGroup", { id: undefined });
```

jaf107: handled in https://github.com/optimizely-axiom/optiaxiom/pull/540

PR #532, opened 2024-11-08
add Spotlight component

radnan:


apps/storybook/src/overlays/Spotlight.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+              K
+            </Kbd>
+          }
+          w="240"
+        >
+          Quick search...
+        </SpotlightTrigger>
+
+        <SpotlightContent>
+          <SpotlightInput placeholder="Languages..." />
```
jaf107, line 127: Does `Attributes` or `Properties` seem better here, instead of `Languages` bhai?
```suggestion
          <SpotlightInput placeholder="Properties..." />
```

apps/storybook/src/overlays/Spotlight.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <Kbd keys="command" variant="subtle">
+              K
+            </Kbd>
+          }
+          w="240"
+        >
+          Quick search...
+        </SpotlightTrigger>
+
+        <SpotlightContent>
```
jaf107, line 126: Bhai, why aren't we passing the `SpotlightInput` in the addOnBefore prop of `Spotlight`

radnan: So addons are slots that let you add/replace things down the chain

A good example is https://github.com/optimizely-axiom/optiaxiom/blob/5f763137e5fbbf44f391571e0040f699ad92bae6/packages/react/src/command-checkbox-item/CommandCheckboxItem.tsx#L20-L29

where we’re adding a checkbox before any other thing in the addon

another example is inputs where we want the addon to go inside the container but next to the actual inputs

and another example is AlertDialogTitle which allows consumers to actually replace the alert icon with something they’d like

And in some cases addons have special styles depending on if they’re placed in the start or end

I suppose we could’ve instead broken these components down into their individual parts and made them really modular - like:

```tsx
<Input>
  <InputAddon />
  <InputControl />
</Input>
```

but that makes the api a bit cumbersome for relatively simple things - hence why we have this addon approach

But in this case with Spotlight we’re just doing pure composition - and there’s no need to create slots.

now that you bring it up - I think we should actually remove the addons from Sidenav as well - since that doesn’t really need slots.

jaf107: Yes bhai. I kind of related this part with Sidenav and as such had this query. Thank you for the descriptive reply!

packages/react/src/spotlight-input/SpotlightInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+import * as styles from "./SpotlightInput.css";
+
+type SpotlightInputProps = ComponentPropsWithoutRef<typeof CommandInput>;
+
+export const SpotlightInput = forwardRef<HTMLInputElement, SpotlightInputProps>(
+  ({ className, ...props }, ref) => {
+    return (
+      <CommandInput
+        addonAfter={
+          <DialogClose fontSize="xs" fontWeight="500" h="xs" size="sm">
```
toha993, line 15: Should we keep the `close` as focusable vai as we are using it in `Input`

packages/react/src/spotlight-content/SpotlightContent.css.ts
```diff
@@ -0,0 +1,13 @@
+import { recipe, style } from "../vanilla-extract";
+
+export const content = recipe({
+  base: [
+    {
+      overflow: "hidden",
+    },
+    style({
+      top: "12vh",
```
toha993, line 9: Vai, should we use use `vh` or `dvh` here?

packages/react/src/combobox-input/ComboboxInput.tsx
```diff
@@ -1,5 +1,30 @@
...
+type ComboboxInputProps = ComponentPropsWithoutRef<typeof CommandInput>;
+
+export const ComboboxInput = forwardRef<HTMLInputElement, ComboboxInputProps>(
+  (props, ref) => {
+    const { downshift, highlightedItem } = useCommandContext("ComboboxInput");
 
-export const ComboboxInput = CommandInput;
+    return (
+      <CommandInput
+        addonBefore={<IconMagnifyingGlass />}
```
toha993, line 15: Handled it in #518 here vai.

apps/storybook/src/overlays/Spotlight.stories.tsx
radnan: addressed

packages/react/src/spotlight-content/SpotlightContent.css.ts
radnan: addressed - also moved all dialogs to top of screen since dialog content can shift and should not change depending on the content (which it does not in main)

and also move alert dialogs to the top third of the screen so it looks better

packages/react/src/combobox-input/ComboboxInput.tsx
radnan: removed

PR #540, opened 2024-11-12
fix `SelectGroupContext` - provide default value

jaf107:
fixes the comment of #519


toha993:
Can we add a changeset? @jaf107

radnan:
I guess since this is a follow up to a feature there’s no need for a changeset - the original should cover everything as we’ve not done a release yet

toha993 (approved):

PR #552, opened 2024-11-13
Add `intent` for Progress component

toha993:


packages/react/src/progress/Progress.css.ts
```diff
@@ -0,0 +1,38 @@
...
+      false: style({
+        borderBottomLeftRadius: theme.borderRadius.full,
+        borderTopLeftRadius: theme.borderRadius.full,
+      }),
+      true: style({
+        borderRadius: theme.borderRadius.full,
+      }),
+    },
+    intent: {
+      default: {
```
radnan, line 25: Should be called primary
```suggestion
      primary: {
```

packages/react/src/progress/Progress.css.ts
```diff
@@ -0,0 +1,38 @@
...
+        bg: "bg.accent",
+      },
+      error: {
```
radnan, line 28: Should be called danger
```suggestion
      danger: {
```

radnan (commented):
I guess we should update the token names to match the intent names used in other components or vice versa - but basically the value of intent should always be same set of values across components

you can always look at badge since it uses all the intent values

packages/react/src/progress/Progress.css.ts
```diff
@@ -0,0 +1,38 @@
...
+  variants: {
+    fullWidth: {
+      false: style({
+        borderBottomLeftRadius: theme.borderRadius.full,
+        borderTopLeftRadius: theme.borderRadius.full,
+      }),
+      true: style({
+        borderRadius: theme.borderRadius.full,
+      }),
+    },
```
radnan, line 23: don't think we need this - since you used `overflow: hidden` on the parent

packages/react/src/progress/Progress.css.ts
```diff
@@ -0,0 +1,38 @@
...
+import { recipe, type RecipeVariants, style } from "../vanilla-extract";
+
+export const progress = recipe({
+  base: [
+    {
+      h: "full",
+    },
+    style({
+      transition: "all 400ms",
+    }),
```
radnan, line 12: why custom style here instead of the transition all sprinkle like before?

apps/storybook/src/feedback/Progress.stories.tsx
```diff
@@ -53,3 +65,59 @@ export const CompletionStages: Story = {
     );
   },
 };
+
+export const AnimatedProgress = () => {
```
radnan, line 69: i don't think we need this story?

packages/react/src/progress/Progress.css.ts
```diff
@@ -0,0 +1,38 @@
+import { theme } from "@optiaxiom/globals";
+
+import { recipe, type RecipeVariants, style } from "../vanilla-extract";
+
+export const progress = recipe({
```
radnan, line 5: let's call this `indicator` since the outer component is the actual progress component/track
```suggestion
export const indicator = recipe({
```

packages/react/src/progress/Progress.css.ts
toha993: The transition didn't work expectedly before, so I added it check it vai. Going to revert it.

radnan (approved):

PR #574, opened 2024-11-20
refactor comboboxes to use composition of Listbox

radnan:


packages/react/src/select-content/SelectContent.tsx
```diff
@@ -18,21 +20,25 @@ export const SelectContent = forwardRef<HTMLDivElement, SelectContentProps>(
...
-              children
-            )}
-          </ul>
-        </SelectContentImpl>
-      </ListboxBase>
+      isOpen && (
+        <Portal asChild>
+          <MenuListbox asChild provider="popper">
+            <SelectContentImpl ref={ref} {...props}>
+              <ul {...downshift.getMenuProps({}, { suppressRefError: true })}>
```
toha993, line 27: Vai, shouldn't it be a `Box`? as the child is also `Box`.

apps/storybook/src/comboboxes/Autocomplete.stories.tsx
```diff
@@ -237,17 +230,11 @@ export const Controlled: Story<(typeof books)[number]> = {
...
-                  key={item.id}
-                >
-                  {item.title}
-                </AutocompleteItem>
-              ))}
-            </AutocompleteList>
+            {items.map((item) => (
+              <AutocompleteItem asChild item={item} key={item.id}>
+                <ListboxRadioItem>{item.title}</ListboxRadioItem>
+              </AutocompleteItem>
```
toha993, line 236: Vai, did we use the `ListboxRadioItem` intentionally here? Can't we use the `AutocompleteRadioItem`?

packages/react/src/listbox-item/ListboxItem.tsx
```diff
@@ -31,11 +31,12 @@ export type ListboxItemBaseProps<
   >
 >;
 
-export const ListboxItemBase = forwardRef<HTMLDivElement, ListboxItemBaseProps>(
+export const ListboxItem = forwardRef<HTMLDivElement, ListboxItemProps>(
```
toha993, line 34: Vai, do we need to set `role="option"` for the `Item` component?

apps/storybook/src/comboboxes/Autocomplete.stories.tsx
radnan: nope - replaced

packages/react/src/listbox-item/ListboxItem.tsx
radnan: no, technically it will have `role="option"` or `role="menuitem"` depending on where it's used and when combined with another component that provides the behavior - but this component is only for appearance purposes and doesn't provide any behavior

PR #579, opened 2024-11-22
refactor dialog header and actions

radnan:


packages/react/src/drawer-trigger/DrawerTrigger.tsx
```diff
@@ -0,0 +1,18 @@
+import * as RadixDialog from "@radix-ui/react-dialog";
```
toha993, line 1: Vai, should we keep it as RadixDrawer like other Drawer components?

packages/react/src/drawer-content/DrawerContent.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Backdrop } from "../backdrop";
+import { type BoxProps } from "../box";
+import { useDrawerContext } from "../drawer-context";
+import { Paper } from "../paper";
+import { Transition } from "../transition";
+import * as styles from "./DrawerContent.css";
+
+type DrawerContentProps = BoxProps<
+  typeof RadixDrawer.Content,
+  NonNullable<styles.DrawerVariants>
```
toha993, line 14: Vai, do we need the NonNullable here?

packages/react/src/dialog-content/DialogContent.tsx
```diff
@@ -50,22 +40,7 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
...
-                      asChild
-                      icon={<IconX />}
-                      size="sm"
-                      {...styles.close()}
-                    >
-                      <RadixDialog.Close />
-                    </Button>
-                  )}
-                </RadixDialog.Content>
+                <RadixDialog.Content ref={ref}>{children}</RadixDialog.Content>
```
toha993, line 43: Vai, when there's no focusable item in the content, opening the drawer has a `black` outline for edge / `blue` outline for chrome.

packages/react/src/drawer-content/DrawerContent.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+            <Transition
+              data-side={
+                mapPositionToTransitionSide[
+                  position as keyof typeof mapPositionToTransitionSide
+                ]
+              }
+              type="fade"
+            >
+              <Paper
```
toha993, line 45: Vai, when there's no focusable item in the content, opening the drawer has a black outline for edge / blue outline for chrome.

packages/react/src/drawer-description/DrawerDescription.tsx
```diff
@@ -0,0 +1,20 @@
+import * as RadixDrawer from "@radix-ui/react-dialog";
+import { forwardRef } from "react";
+
+import { Text, type TextProps } from "../text";
+import * as styles from "./DrawerDescription.css";
+
+type DrawerDescriptionProps = TextProps<typeof RadixDrawer.Title>;
```
toha993, line 7:
```suggestion
type DrawerDescriptionProps = TextProps<typeof RadixDrawer.Description>;
```

apps/storybook/src/overlays/Drawer.stories.tsx
```diff
@@ -1,63 +1,72 @@
...
+    withFooter,
+    ...args
+  }) {
     return (
-      <Flex>
-        <Button onClick={onOpen}>Open Drawer</Button>
+      <Drawer {...args}>
+        <DrawerTrigger>Open Drawer</DrawerTrigger>
+
+        <DrawerContent {...args}>
```
toha993, line 42: Vai, when we don't pass any description, it will show a warning, we handled this in dialog like this way.

```suggestion
        <DrawerContent 
        {...(!description && { ["aria-describedby"]: undefined })}
         {...args}
         >
```

packages/react/src/drawer-trigger/DrawerTrigger.tsx
radnan: No, why did we use RadixDrawer in other places? The package and component is dialog and should be called that to prevent confusion.

packages/react/src/drawer-content/DrawerContent.tsx
radnan: Yes - recipe variants can be omitted - meaning they can be `undefined` instead of an object. but that’ll cause issues when merging props - as the final props type can become `any`

packages/react/src/drawer-content/DrawerContent.tsx
radnan: That is expected since dialog will focus trap you inside drawer

but you’ll always have something focusable in the content - at the very least you’ll have a close button. So we should be good.

packages/react/src/drawer-trigger/DrawerTrigger.tsx
toha993: Then, we need to change where we declared this as `RadixDrawer` vai. Previously maintained the alias, as per our component naming.

packages/react/src/drawer-content/DrawerContent.tsx
toha993: Got it vai.

packages/react/src/drawer-content/DrawerContent.tsx
toha993: Then, can we export the type with NonNullable like `ButtonBase.css.ts`, `Alert.css.ts` for consistency over components?

packages/react/src/dialog-description/DialogDescription.tsx
```diff
@@ -15,17 +16,9 @@ type DialogDescriptionProps = BoxProps<
 export const DialogDescription = forwardRef<
   HTMLParagraphElement,
```
toha993, line 17: Vai, should it be a `HTMLParagraphElement` or `HTMLDivElement`?

toha993: For drawer with the same structure, we used the HTMLDivElement, wasn't sure about it vai.

packages/react/src/drawer-trigger/DrawerTrigger.tsx
radnan: addressed

PR #582, opened 2024-11-25
Fix `Spotlight` story

toha993:


apps/storybook/src/overlays/Spotlight.stories.tsx
```diff
@@ -214,7 +214,13 @@ export const Basic: Story<{ items?: Item[] } & Item> = {
                         </Box>
                       </SpotlightEmpty>
 
-                      <SpotlightItem item={item}>
+                      <SpotlightItem
```
radnan, line 217: We can actually replace this with SpotlightRadioItem - missed replacing this in last PR

toha993: I also thought of using it vai, but wasn't sure if it was okay to use RadioItem.

radnan (approved):

PR #601, opened 2024-12-03
Fix ListboxItem styling

toha993:


packages/react/src/listbox-item/ListboxItem.css.ts
```diff
@@ -73,6 +73,10 @@ export const item = recipe({
 
 export const title = recipe({
   base: [
+    {
+      alignItems: "center",
```
radnan, line 77: we need `flexDirection: row` here if we want to re-use `Flex` component

maybe add a story so it's visually clear what's happening here

toha993: Addressed vai.

apps/storybook/src/comboboxes/Listbox.stories.tsx
```diff
@@ -184,6 +184,21 @@ export const IconMultiSelect: Story = {
   },
 };
 
+export const ItemWithIcon: Story = {
```
radnan, line 187: let's rename this to clarify that this is just rendering icons inside the content/children
```suggestion
export const ComplexContent: Story = {
```

PR #607, opened 2024-12-05
Add maxW to tooltip

toha993:


packages/react/src/tooltip/Tooltip.tsx
```diff
@@ -54,7 +54,9 @@ export const Tooltip = forwardRef<HTMLButtonElement, TooltipProps>(
         </TooltipTrigger>
 
         {!disabled && !empty && (
-          <TooltipContent {...props}>{content}</TooltipContent>
+          <TooltipContent maxW="xs" {...props}>
```
radnan, line 57: can we add a `LongContent` story for this

radnan: something really simple like:

```tsx

export const LongContent: Story = {
  args: {
    children: <Button>Hover</Button>,
    content:
      "This is a really long piece of tooltip content to check how the text is wrapped.",
    defaultOpen: true,
  },
};
```

toha993: Sure vai

radnan (approved):

PR #631, opened 2024-12-12
Add `FieldLabel` component

toha993:


packages/react/src/field-label/FieldLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+  "div",
+  {
+    info?: ReactNode;
+    inputId?: string;
+    label: ReactNode;
+    required?: boolean;
+  }
+>;
+
+export const FieldLabel = forwardRef<HTMLDivElement, FieldLabelProps>(
```
radnan, line 25: we also need to re-use this FieldLabel ourselves in Field component

PR #674, opened 2025-01-05
Add hover support for popover

toha993:


radnan (commented):
Let’s hold off on this for now

and wherever we need hover just use a tooltip in v1

Adding hover here won’t be so easy - there’s a lot of things we need to consider

toha993:
Not planned.

PR #712, opened 2025-01-22
Add `DropdownMenuCheckboxItem` component

toha993:


packages/react/src/dropdown-menu-checkbox-item/DropdownMenuCheckboxItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const DropdownMenuCheckboxItem = forwardRef<
+  HTMLDivElement,
+  DropdownMenuCheckboxItemProps
+>(
+  (
+    { asChild, checked, children, onCheckedChange, onSelect, ...props },
+    ref,
+  ) => {
+    return (
+      <ListboxRadioItem
```
radnan, line 20: need to replace with `ListboxCheckboxItem`

toha993: Addressed vai.

radnan (approved):

PR #723, opened 2025-01-26
Add `DateInput` component

toha993:


.changeset/twelve-fans-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/react": minor
```
radnan, line 2: Should be patch

apps/storybook/src/components/DateInput.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/qs72V79n1s9wYOcZ1TzBwM/Components-V2?node-id=5743-27400",
+    },
+  },
+} as Meta<typeof DateInput>;
+
+type Story = StoryObj<typeof DateInput>;
+
+export const Default: Story = {
```
radnan, line 17: Our convention is Basic for the first story name

apps/storybook/src/components/DateInput.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+  args: {
+    placeholder: "Select date",
+  },
+};
+
+export const WithDefaultValue: Story = {
```
radnan, line 23: The “With” prefix is not needed here - just follow the Input story names

.changeset/twelve-fans-visit.md
toha993: Okay vai

packages/react/src/box/Box.css.ts
```diff
@@ -97,7 +97,7 @@ export const box = recipe({
            * 1. Correct the odd appearance in Chrome and Safari.
            * 2. Correct the outline style in Safari.
            */
-          "&:is([type=search])": {
+          "&:is([type=search], [type=date])": {
```
radnan, line 100: Why do we need this here?

apps/storybook/src/components/DateInput.stories.tsx
toha993: Forgot to rename it properly after a change.

packages/react/src/calendar/Calendar.tsx
```diff
@@ -16,6 +16,14 @@ import { CalendarWeekday } from "./CalendarWeekday";
...
   "div",
   {
+    /**
+     * When a selection mode is set, DayPicker will focus the first selected day
+     * (if set) or today's date (if not disabled).
+     *
+     * Use this prop when you need to focus DayPicker after a user action, for
+     * improved accessibility.
+     */
+    autoFocus?: boolean;
```
radnan, line 26: I don’t think we need this to be a prop

just default it to true inside the calendar body

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    const handleDateChange = (date: Date) => {
+      const year = date.getFullYear();
+      const month = date.getMonth();
+      const day = date.getDate();
+      const utcDate = new Date(Date.UTC(year, month, day));
+
+      const formattedDate = utcDate.toISOString().split("T")[0];
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      onChange?.({ target: { value: formattedDate } } as any);
```
radnan, line 37: We can simplify this to just `value: date.toISOString()` ?

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+type DateProps = ComponentPropsWithRef<typeof Input>;
+
+export const DateInput = forwardRef<HTMLInputElement, DateProps>(
+  (
+    { defaultValue, disabled, max, min, onChange, value: valueProp, ...props },
+    ref,
+  ) => {
+    const [open, setOpen] = useState(false);
+    const [innerValue, setValue] = useState(defaultValue);
```
radnan, line 20: Let’s use useControllableState to mirror the value of input

similar to how we do it in Textarea:

https://github.com/optimizely-axiom/optiaxiom/blob/d0c7bc41d8fb8949bc003907270264a6975b108f/packages/react/src/textarea-autosize/TextareaAutosize.tsx#L29

packages/react/src/date-input/DateInput.tsx
radnan, line 37: Can we re-use forceValueChange here?

https://github.com/optimizely-axiom/optiaxiom/blob/d0c7bc41d8fb8949bc003907270264a6975b108f/packages/react/src/search-input/SearchInput.tsx#L49

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+      <Popover onOpenChange={setOpen} open={open}>
+        <PopoverAnchor>
+          <Input
+            addonAfter={
+              <PopoverTrigger
+                appearance="subtle"
+                aria-label="toggle calendar"
+                cursor="default"
```
radnan, line 49: Shouldn’t have this - pointer cursor should be used to denote interactive element

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+                disabled={disabled}
+                icon={<IconAngleDown />}
```
radnan, line 51: We shouldn’t have two addons. Just one with calendar that is the popover trigger

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+              onChange?.(event);
+              setValue(event.target.value);
+            }}
+            onKeyDown={(e) => {
+              if (e.key === " ") {
+                e.preventDefault();
+                setOpen(true);
+              }
+            }}
+            ref={ref}
```
radnan, line 68: Can we use useImperativeHandle https://react.dev/reference/react/useImperativeHandle to override the showPicker https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker  method?

so if some has a ref to the input and invokes `ref.showPicker()` it should open the popover

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+          />
+        </PopoverAnchor>
+        <PopoverContent p="0">
+          <Calendar
+            autoFocus
+            defaultValue={dateValue}
+            max={maxDate}
+            min={minDate}
+            onValueChange={(date) => {
+              handleDateChange(date);
```
radnan, line 81: Let’s in-line this method if not used anywhere else

packages/react/src/box/Box.css.ts
toha993: Followed the pattern we had for search. Removed it vai.

packages/react/src/date-input/DateInput.tsx
toha993: Used the function vai.

packages/react/src/date-input/DateInput.tsx
toha993: I used it, as for disabled, button had `cursor: not-allowed`, but for input that wasn't given. So, when I hover to the trigger, it shows different cursor.

packages/react/src/date-input/DateInput.tsx
toha993: Addressed vai, but it doesn't match with the figma design. The calendar icon comes in the left.

packages/react/src/date-input/DateInput.tsx
toha993: Should we do the same for `SearchInput` also?

packages/react/src/date-input/DateInput.tsx
toha993: Faced this when doing this 
image

Another thing about UtcDate, As I am in GMT+6 location, whenever I select a date, it always saves a day before, so I took the UTC date.

packages/react/src/date-input/DateInput.tsx
toha993: Addressed vai.

apps/storybook/src/components/DateInput.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Button, Flex } from "@optiaxiom/react";
+import { DateInput } from "@optiaxiom/react/unstable";
+import { useRef } from "react";
+
+export default {
+  component: DateInput,
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/qs72V79n1s9wYOcZ1TzBwM/Components-V2?node-id=5743-27575",
```
radnan, line 12:
```suggestion
      url: "https://www.figma.com/design/qs72V79n1s9wYOcZ1TzBwM/Components-V2?node-id=5743:27573",
```

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+} from "react";
+
+import { Calendar } from "../calendar";
+import { IconCalendar } from "../icons/IconCalendar";
+import { Input } from "../input";
+import { Popover } from "../popover";
+import { PopoverAnchor } from "../popover-anchor";
+import { PopoverContent } from "../popover-content";
+import { PopoverTrigger } from "../popover-trigger";
+import { forceValueChange } from "../search-input";
```
radnan, line 18: let's copy paste this function or move it to the `utils` folder

packages/react/src/date-input/DateInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+            disabled={disabled}
+            max={max}
+            min={min}
+            onChange={(event) => {
+              onChange?.(event);
+              setValue(event.target.value);
+            }}
+            onClick={(e) => {
+              e.preventDefault();
+            }}
```
radnan, line 85: don't think we need this?

packages/react/src/date-input/DateInput.tsx
toha993: Okay vai.

packages/react/src/date-input/DateInput.tsx
radnan: oh right forgot how awful dates are in JS

packages/react/src/date-input/DateInput.tsx
radnan: i think so - let's do it in a separate PR tho

packages/react/src/date-input/DateInput.tsx
radnan: interesting - i guess we need it for `focus` as well then?

toha993: No vai, it opens when clicked over day/month/year section.

packages/react/src/date-input/DateInput.tsx
```diff
@@ -48,6 +48,9 @@ export const DateInput = forwardRef<HTMLInputElement, DateInputProps>(
     const minDate = min ? new Date(min) : undefined;
 
     const handleDateChange = (date: Date) => {
+      // calendar deselects the selected date if clicked the selected date again.
+      if (!date) return;
```
radnan, line 52: this is not the right way to achieve this - cause we could add a "reset" button to the calendar which would reset the date

better to disallow toggling date in the calendar using the `required` prop

PR #724, opened 2025-01-27
fix docs and slightly refactor cards

radnan:


packages/react/src/date-input/DateInput.tsx
```diff
@@ -90,17 +73,31 @@ export const DateInput = forwardRef<HTMLInputElement, DateInputProps>(
               }
             }}
             ref={ref}
-            value={value}
-            {...props}
             type="date"
+            {...props}
```
toha993, line 77: Vai, I was thinking about this approach, if we pass type before props, anyone can just change the type anytime, so it won't be date-input anymore. Do we want to keep that option open or not being able to change the type anyway.

radnan: It’s fine we can keep it open - no need to limit it artificially here

the other type they could use is `datetime-local` I guess. But there’s no need to limit it from our side.

PR #751, opened 2025-02-06
add `labelPosition` & remove `description`  props support in `Switch`

jaf107:
resolves : https://github.com/optimizely-axiom/optiaxiom/issues/747


jaf107:
the label support can be easily achieved using 
`flex-direction: row-reverse`

So, closing this PR

PR #831, opened 2025-03-04
add DatePicker and DateRangePicker components

radnan:


packages/react/src/date-picker-content/DatePickerContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const DatePickerContent = forwardRef<
+  HTMLInputElement,
+  DatePickerContentProps
+>(({ children, className, holiday, weekend, ...props }, ref) => {
+  const { setValue, value } = useDatePickerContext("DatePickerContent");
+
+  return (
+    <PopoverContent ref={ref} {...styles.content({}, className)} {...props}>
+      <Calendar
```
toha993, line 19: Should we keep the clear selection by default like `DateInput` vai?

radnan: i'm not sure - date pickers are quite flexible and should not have too much defaults

users are free to inject any content via `children` - so left it out for now

packages/react/src/calendar/Calendar.tsx
```diff
@@ -95,13 +119,14 @@ export const Calendar = forwardRef<HTMLDivElement, CalendarProps>(
...
-          defaultMonth={value}
+          defaultMonth={value && "from" in value ? value.from : value}
           endMonth={max}
-          mode="single"
+          mode={mode as "single"}
           modifiers={{ holiday, weekend }}
-          onSelect={setValue}
+          numberOfMonths={mode === "range" ? 2 : 1}
+          onSelect={setValue as (value: Date | undefined) => void}
           required
```
toha993, line 128: We can't change the start date forward after the first selection, we can only change the start date backward.
It might be related to the required prop.
https://daypicker.dev/docs/selection-modes#range-mode

toha993 (approved):

PR #889, opened 2025-04-06
Update theme import path from @optiaxiom/react to @optiaxiom/globals

toha993:


toha993:
We export `createToaster, theme, toaster, tokens } from "@optiaxiom/globals"` internally from `@optiaxiom/react`

PR #910, opened 2025-04-18
Support truncation in `Breadcrumb` items

toha993:


apps/storybook/src/components/Breadcrumb.stories.tsx
```diff
@@ -99,3 +99,25 @@ export const WithDropdown: Story = {
     ),
   },
 };
+
+export const Truncated: Story = {
+  args: {
+    children: (
+      <>
+        <BreadcrumbItem>
+          <BreadcrumbLink href="/">
```
radnan, line 108: i think we can just remove the height and display styles from `BreadcrumbLink`. there's no need for them.

radnan: there's quite a few redundant styles there now - we can also remove `rounded` - not sure why that's there on breadcrumb link

radnan (approved):

PR #1081, opened 2025-06-19
add Resize support for  `Aside`

jaf107:


packages/react/src/aside/Aside.css.ts
```diff
@@ -0,0 +1,16 @@
...
+export const root = recipe({
+  base: [
+    {
+      color: "fg.default",
+      display: "flex",
+      flexDirection: "column",
+      h: "full",
+    },
+    style({
+      boxShadow:
```
toha993, line 12: Don't we have token for it? Should we use boxShadow which isn't under design token?

packages/react/src/aside/Aside.tsx
```diff
@@ -0,0 +1,18 @@
+import { forwardRef } from "react";
+
+import { Box, type BoxProps } from "../box";
+import * as styles from "./Aside.css";
+
+export type Aside = BoxProps<"div">;
```
toha993, line 6:
```suggestion
export type AsideProps = BoxProps<"div">;
```

packages/react/src/aside/AsideFooter.css.ts
```diff
@@ -0,0 +1,15 @@
...
+      borderColor: "border.secondary",
+      borderT: "1",
+      flexDirection: "row",
+      gap: "16",
+      justifyContent: "flex-end",
+      px: "24",
+      py: "20",
+    },
+  ],
+});
```
toha993, line 15: Let's take it to inline sprinkles

PR #1160, opened 2025-08-08
allow keyboard date input from range picker

### LINKED ISSUE #1159: Make it easier to select dates in date range picker
Date range picker only allows selecting dates from calendar - unlike DateInput which allows manually entering dates

So navigating to far away dates is very cumbersome

radnan:
resolves #1159


radnan:
closing in favor of #1163

Issue #1227, opened 2025-11-06
add range component


jaf107:
Props Support - https://www.radix-ui.com/primitives/docs/components/slider

```
<Box {...boxProps}>                          // Wrapper for spacing props
  <SliderPrimitive.Root {...styles.root()}>  // Root: 20px height container
    <Box asChild {...styles.track()}>        // Track: 8px gray background bar
      <SliderPrimitive.Track>
        <Box asChild {...styles.range()}>    // Range: Blue filled portion
          <SliderPrimitive.Range />
        </Box>
      </SliderPrimitive.Track>
    </Box>
    
    <Box asChild {...styles.thumb()}>        // Thumb: 20px draggable circle
      <SliderPrimitive.Thumb />
    </Box>
  </SliderPrimitive.Root>
</Box>
```

### Usage
```
<Range 
  defaultValue={[25]}
  min={0}
  max={100}
  step={5}
  w="384"
/>
```

PR #1228, opened 2025-11-06
add `Range` component

jaf107:
resolves https://github.com/optimizely-axiom/optiaxiom/issues/1227


packages/react/src/index.ts
```diff
@@ -40,6 +40,7 @@ export * from "./paper";
 export * from "./popover";
 export * from "./progress";
 export * from "./radio";
+export * from "./range";
```
radnan: move to unstable for now

PR #1230, opened 2025-11-17
Add disabled pill support for PillMenu

toha993:
#1229


dustyRAIN (commented):
should we add the dimisss functionality (cross button to deselect pill option) to the pillmenu?

PR #1251, opened 2025-12-03
Update interaction test for `pillmenu`

toha993:


apps/storybook/src/components/PillMenu.stories.tsx
```diff
@@ -99,28 +99,19 @@ export const Disabled: Story = {
   play: async () => {
```
radnan: looks like UI test is still failing - let's just comment this out and re-visit later

PR #1352, opened 2026-02-19
Add Block.Image block type 

jaf107:
Add support for rendering images in block documents.
                                                                              
  - Add `Image` base component                              
  - Add `Block.Image` to block document spec and renderer
  - Support single image and multi-image grid layouts via `Block.Group`

### DIFF
--- a/packages/react/src/block-document/BlockImage.tsx
+++ b/packages/react/src/block-document/BlockImage.tsx
@@ -16,7 +16,7 @@ export type BlockImageProps = BoxProps<
 
 export function BlockImage(props: BlockImageProps) {
   return (
-    <Box asChild display="block" objectFit="cover" rounded="lg" {...props}>
+    <Box asChild display="block" objectFit="cover" {...props}>
       <img />
     </Box>
   );


packages/react/src/index.ts
```diff
@@ -26,6 +26,7 @@ export * from "./flex";
 export * from "./grid";
 export * from "./group";
 export * from "./heading";
+export * from "./image";
```
radnan: do not export this Image component

packages/web-components/package.json
```diff
@@ -81,6 +81,7 @@
     "./Grid": "./dist/components/Grid.js",
     "./Group": "./dist/components/Group.js",
     "./Heading": "./dist/components/Heading.js",
+    "./Image": "./dist/components/Image.js",
```
radnan: do not export this component

packages/react/src/block-document/BlockImage.tsx
```diff
@@ -0,0 +1,9 @@
+import type { BlockImageProps } from "./schemas";
+
+import { Image } from "../image";
+
+export function BlockImage(props: BlockImageProps) {
+  return <Image display="block" objectFit="cover" rounded="md" {...props} />;
```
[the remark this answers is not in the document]
jaf107: It looks like it has `borderRadius` of 8px (`lg`) by default
ref: https://www.figma.com/design/tt1XbZP9uV2glsyiFMNUg5/Adaptive-Block?node-id=2596-25665&m=dev

PR #1471, opened 2026-04-07
feat: auto-focus and escape key improvements in Proteus

toha993:
### 1. Auto-focus first interactive element in Proteus documents

On mount, the first focusable element (input, select, textarea, or custom component with `tabIndex >= 0`) is automatically focused when `blocking: true` is set on the document.

**Approach:** Uses `TreeWalker` in `ProteusDocumentShell`'s mount effect, gated on `element.blocking`. This is the same logic Radix uses internally in `@radix-ui/react-focus-scope` (`getTabbableCandidates`). Only runs for blocking documents to avoid focusing non-interactive content.

**Why this approach over alternatives:**
- **`FocusScope` from Radix** — Falls back to focusing the container div when no tabbable elements exist, causing visible focus rings on text documents. Also couldn't scope to just the body content without wrapping in a div that broke flex gap layout.
- **`autoFocus` HTML attribute** — Setting `autoFocus` on inputs inside a `Map` template applies to all inputs; browser focuses the last one.
- **Data-driven `autoFocus`** — Leaks UI concerns into the data model.
- **No `blocking` gate** — Caused focus rings on plain text/image documents.

**Special case for `ProteusQuestion`:** Has its own focus handling in a single `useEffect` because pagination buttons (Previous/Next/Dismiss) precede options in DOM order. On initial mount (`else` branch), it focuses the first option. On index change (`if` branch), it animates and focuses.

### 2. Fix escape key behavior in Question component

Previously, focusing an option via keyboard navigation auto-selected it (via `onFocus` handler on `single_select`), so pressing Escape required two presses: first to clear the focus-driven selection, then to skip.

Now tracks explicit user interaction (`interactedRef`) — click, Enter/Space on options, or typing in "Something else" input. First Escape skips directly when no real interaction occurred. `interactedRef` resets when navigating between questions.

## Test plan
- [ ] AskAgentInput: first text input receives focus on mount
- [ ] AskUserQuestion: first option receives focus (not pagination buttons)
- [ ] Basic story (text only): no visible focus ring
- [ ] FormWithInputs (no blocking): no autofocus
- [ ] AskUserQuestion: Escape without interaction skips the question
- [ ] AskUserQuestion: after clicking an option, Escape clears selection first
- [ ] AskUserQuestion: after navigating to Q2, Escape skips (interactedRef reset)
- [ ] AskUserQuestion: typing in "Something else" counts as interaction

### DIFF
--- a/packages/proteus/src/proteus-question/ProteusQuestion.tsx
+++ b/packages/proteus/src/proteus-question/ProteusQuestion.tsx
@@ -117,7 +117,7 @@ export function ProteusQuestion({ questions }: ProteusQuestionProps) {
       onKeyDown={(event) => {
         if (event.key === "Escape") {
           event.preventDefault();
-          if (valid) {
+          if (type === "multi_select" && valid) {
             onValueChange(null);
           } else {
             onSkip();
--- a/packages/proteus/src/proteus-question/ProteusQuestion.tsx
+++ b/packages/proteus/src/proteus-question/ProteusQuestion.tsx
@@ -60,11 +60,11 @@ export function ProteusQuestion({ questions }: ProteusQuestionProps) {
         },
       );
       lastIndexRef.current = currentIndex;
-      const item =
-        questionRef.current?.querySelector<HTMLElement>("[data-selected]") ??
-        questionRef.current?.querySelector<HTMLElement>("[tabindex]");
-      item?.focus();
     }
+    const item =
+      questionRef.current?.querySelector<HTMLElement>("[data-selected]") ??
+      questionRef.current?.querySelector<HTMLElement>("[tabindex]");
+    item?.focus();
   }, [currentIndex]);
 
   const otherInputRef = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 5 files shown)


packages/proteus/src/proteus-question/ProteusQuestion.tsx
```diff
@@ -117,8 +119,9 @@ export function ProteusQuestion({ questions }: ProteusQuestionProps) {
       onKeyDown={(event) => {
         if (event.key === "Escape") {
           event.preventDefault();
-          if (valid) {
+          if (valid && interactedRef.current) {
```
radnan: let's not do this - manually maintaining `interactedRef.current` isn't good

i thought we talked about the approach being if it's a single_select question then we immediately reset and skip to the next question? whereas for multi_select we will either reset or skip

toha993: I missed the approach vai while implementing - correcting it.

toha993: Addressed vai.

radnan (approved):

PR #1529, opened 2026-04-23
fix: truncate toast title text to prevent overflow

jaf107:
## Summary

When a toast notification displays very long text (e.g., a 300+ character string with no word breaks), the toast grows unbounded vertically, creating an oversized notification.

This adds `lineClamp="2"` and `overflow-wrap: break-word` to `ToastTitle` so that long text is clamped to 2 lines with an ellipsis. Normal-length messages that naturally span 2 lines (like the existing `LongContent` story) are unaffected.

### Before
image

### After
image

## Changes

- **`ToastTitle.tsx`**: Added `lineClamp="2"` as a default prop and applied CSS recipe for word-breaking
- **`ToastTitle.css.ts`** (new): CSS recipe with `overflowWrap: "break-word"` for long unbreakable strings

## Context

Reported via CMP internal ticket CCP-16762 — a field with a very long name caused the toast to display an oversized notification when the field was removed.

### DIFF
--- a/packages/react/src/toast/ToastTitle.tsx
+++ b/packages/react/src/toast/ToastTitle.tsx
@@ -2,6 +2,7 @@ import * as RadixToast from "@radix-ui/react-toast";
 import { forwardRef } from "react";
 
 import { Text, type TextProps } from "../text";
+import * as styles from "./ToastTitle.css";
 
 export type ToastTitleProps = TextProps<typeof RadixToast.Title>;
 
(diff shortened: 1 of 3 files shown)


MunifTanjim (approved):

packages/react/src/toast/ToastTitle.tsx
```diff
@@ -10,9 +11,9 @@ export type ToastTitleProps = TextProps<typeof RadixToast.Title>;
  * @extends Text
  */
 export const ToastTitle = forwardRef<HTMLDivElement, ToastTitleProps>(
-  ({ children, ...props }, ref) => {
+  ({ children, className, ...props }, ref) => {
     return (
-      <Text asChild flex="1" {...props}>
+      <Text asChild lineClamp="2" {...styles.title({}, className)} {...props}>
```
tanvirstreame, line 16: Just like you moved flex to the css.ts file, wouldn’t it be better to put lineClamp there too?

jaf107: Bhai, `lineClamp` is a variant on the Text recipe ([Text.css.ts](https://github.com/optimizely-axiom/optiaxiom/blob/b4595355e12d6b24200d013d4083db8ad84e9383/packages/react/src/text/Text.css.ts#L21)), not a sprinkle prop like flex. It expands to multiple CSS properties (overflow: hidden, text-overflow: ellipsis, display: -webkit-box, -webkit-box-orient: vertical, -webkit-line-clamp: 2).

For this reason it can only be passed as a JSX prop to <Text> and it can't be moved into our recipe's style({}) block like `flex`.

PR #1547, opened 2026-04-30
fix(proteus): fix required on ProteusFileUpload

vk1cd314:
## Problem

`ProteusFileUpload` declared a `required` prop but never read it. The surrounding form's submit button stayed enabled even when a required file param was unfilled, and the structured-message resolver dropped the empty file from `files`, leaving the host with no signal a required input was missing.

## Fix

- `ProteusFileUpload` renders a `VisuallyHidden` controlled `<input>` whose value reflects upload completion and carries the `required` attribute, so native `checkValidity()` has something to gate on.
- `ProteusDocumentShell` re-runs `checkValidity()` whenever `data` changes, so submit re-enables once the uploaded URL is written into form data.

### DIFF
--- a/packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
+++ b/packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
@@ -5,9 +5,11 @@ import {
   FileUploadList,
   type FileUploadListProps,
   FileUploadTrigger,
+  VisuallyHidden,
 } from "@optiaxiom/react/unstable";
-import { useCallback, useState } from "react";
+import { useCallback, useRef, useState } from "react";
 
+import { useObserveValue } from "../hooks";
 import { useProteusDocumentContext } from "../proteus-document/ProteusDocumentContext";
 import { useProteusDocumentPathContext } from "../proteus-document/ProteusDocumentPathContext";
 
@@ -31,7 +33,11 @@ export type ProteusFileUploadProps = {
 
 type Item = FileUploadListProps["items"][number];
 
-export function ProteusFileUpload({ accept, name }: ProteusFileUploadProps) {
+export function ProteusFileUpload({
+  accept,
+  name,
+  required,
+}: ProteusFileUploadProps) {
   const { onDataChange, onUpload, readOnly } = useProteusDocumentContext(
     "@optiaxiom/proteus/ProteusFileUpload",
   );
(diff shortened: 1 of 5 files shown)


packages/proteus/src/proteus-document/ProteusDocumentShell.tsx
```diff
@@ -127,7 +127,7 @@ export function ProteusDocumentShell({
     if (formRef.current) {
       setValid(formRef.current.checkValidity());
     }
-  }, []);
+  }, [data]);
```
radnan: why is this needed?

packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
```diff
@@ -77,12 +82,23 @@ export function ProteusFileUpload({ accept, name }: ProteusFileUploadProps) {
...
       disabled={!onUpload || readOnly}
       onFilesDrop={handleFilesDrop}
     >
+      <VisuallyHidden asChild>
+        <input
+          aria-hidden
+          onChange={() => {}}
+          required={required}
+          tabIndex={-1}
+          value={isFilled ? "1" : ""}
```
radnan: also set the `name`?
```suggestion
          value={isFilled ? "1" : ""}
          name={name}
```

packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
```diff
@@ -77,12 +82,23 @@ export function ProteusFileUpload({ accept, name }: ProteusFileUploadProps) {
...
+
   return (
     <FileUpload
       accept={accept}
       disabled={!onUpload || readOnly}
       onFilesDrop={handleFilesDrop}
     >
+      <VisuallyHidden asChild>
+        <input
+          aria-hidden
```
radnan: just use type="hidden" ?

probly don't need a VisuallyHidden in that case either

packages/proteus/src/proteus-document/ProteusDocumentShell.tsx
vk1cd314: would need to recheck validity everytime the form data changes, was having an issue where the agent run button was never enabled so had to do this.

packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
vk1cd314: So I initially did try that and it didn't work, i found out that type="hidden" is exempt from HTML5 constraint validation, so that's when I changed it to VisuallyHidden.

radnan (approved):

PR #1556, opened 2026-05-04
feat(proteus): return file upload metadata object instead of URL string

vk1cd314:
Change `UploadFile` to return `Promise<FileUploadMetadata>` (`{ name, link }`) instead of a URL string, and make both `UploadFile<F>` and `StructuredMessage<F>` generic with that base as the default. Hosts can narrow the metadata to their own richer type (e.g. `<UploadFile<FileMetaData>>`) and `message.files` will be typed accordingly without proteus needing to know the consumer's schema.

**Breaking** for hosts implementing `onUpload`.

### DIFF
--- a/packages/proteus/src/proteus-document/ProteusDocumentContext.ts
+++ b/packages/proteus/src/proteus-document/ProteusDocumentContext.ts
@@ -5,8 +5,9 @@ import { createContext } from "@radix-ui/react-context";
 import type { ProteusEventHandler } from "./schemas";
 
 export type FileUploadMetadata = {
+  [key: string]: unknown;
+  link: string;
   name: string;
-  url: string;
 };
 
 export type UploadFile = (file: File) => Promise<FileUploadMetadata>;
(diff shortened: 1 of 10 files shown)


packages/proteus/src/proteus-document/ProteusDocumentContext.ts
```diff
@@ -4,7 +4,12 @@ import { createContext } from "@radix-ui/react-context";
 
 import type { ProteusEventHandler } from "./schemas";
 
-export type UploadFile = (file: File) => Promise<string>;
+export type FileUploadMetadata = {
+  name: string;
+  url: string;
```
radnan: is this correct? should be something like
```suggestion
  file_id: string;
  mime_type: string;
  name: string;
  size: number;
  uri: string;
  link: string;
```

vk1cd314: Oh, you wanted all the information from the filemetadata, okay changing

radnan (approved):

PR #1557, opened 2026-05-05
Add manual trigger

toha993:
## Summary

Adds a `manualSuggestion` boolean to `unstable_SurfaceProvider`. When set, an Axiom `Input`/`Textarea` no longer auto-fetches a suggestion on every keystroke — the user clicks an Opal trigger button to ask, and the resulting value-suggestion is auto-accepted silently.

## Problem

Today, every keystroke in `Input`/`Textarea` fires a debounced `surface.track({ name: "changed", ... })`, which the host listens to and uses to fetch a suggestion. The suggestion arrives on the surface and `SuggestionPopover` shows the Opal avatar with a popover-Accept flow. For some flows, the host wants the user — not value-change debounce — to ask for a suggestion, and to accept it without an extra confirmation step.

## Public API

`SurfaceContextValue` (in `@optiaxiom/globals`) gains one optional flag:

```ts
manualSuggestion?: boolean;
```

Host wires it on `unstable_SurfaceProvider`:

```tsx
<unstable_SurfaceProvider
  ...existing props (accept, reject, executeTool, track, suggestions, ...)
  manualSuggestion
>
  <Input value={value} onChange={onChange} />
  <Textarea value={value} onChange={onChange} />
</unstable_SurfaceProvider>
```

No new props on `Input`, `Textarea`, or `SuggestionPopover`.

## Behavior

**\`manualSuggestion\` falsy (default).** Unchanged — keystrokes fire \`track({ changed })\` after 300ms debounce; host produces a suggestion; popover Opal avatar appears; user clicks it, then Accept.

**\`manualSuggestion === true\`.** \`InputControl\` skips the keystroke \`debouncedTrack(newValue)\` call. \`SuggestionPopover\` reads \`surface.manualSuggestion\` and renders the manual branch — a plain Button (Opal avatar, no Popover wrapper) with these states:

| State | Trigger | UI |
|---|---|---|
| idle | no request in flight | enabled Opal avatar button |
| pending | user clicked, awaiting suggestion | disabled, Button `loading` spinner overlays the Opal avatar |
| arriving | suggestion appears on surface while pending | auto-call `surface.accept(id)`; clear pending; button stays visible |

Click handler in idle state:
1. Set local \`pending = true\`.
2. Call \`surface.track({ name: "requested", value: surface.value })\` — a new track type distinct from the keystroke-debounced \`changed\`.

The host's \`track\` handler produces a suggestion and pushes it via the provider's \`suggestions\` prop. The component picks the first entry (\`suggestions[0]\`) and auto-accepts; the button remains rendered so the user can request another suggestion at any time.

## File-level changes

- **\`packages/globals/src/surface-context.ts\`** — add optional \`manualSuggestion\` to \`SurfaceContextValue\`. Provider already spreads remaining props onto context value.
- **\`packages/react/src/input/InputControl.tsx\`** — wrap the \`debouncedTrack(newValue)\` call in \`if (!surface?.manualSuggestion) { ... }\`.
- **\`packages/react/src/suggestion/SuggestionPopover.tsx\`** — small router that reads \`surface.manualSuggestion\` and dispatches.
- **\`packages/react/src/suggestion/SuggestionAutoPopover.tsx\`** (new) — existing popover-with-Accept behavior, extracted.
- **\`packages/react/src/suggestion/SuggestionManualTrigger.tsx\`** (new) — \`Button\` with Opal avatar icon and built-in \`loading\` spinner; auto-accepts the first arriving suggestion; emits \`track({ requested })\` on click.
- **Stories.** \`Input\` and \`Textarea\` \`WithManualSuggestion\` — \`SurfaceProvider\` with \`manualSuggestion\`, \`track\` handler that pushes a suggestion 1.2s after a \`requested\` event.

## Edge cases

- **No suggestion arrives.** \`pending\` stays true; button stays disabled. Out of scope for v1; host responsible for resolving (e.g., push a no-op suggestion).
- **Suggestion matches current value.** Picking \`suggestions[0]\` (rather than filtering by \`value !== surface.value\`) ensures we always accept and clear pending — avoids a perpetual pending state.
- **Value changes while pending.** Pending state preserved; arriving suggestion still auto-accepts and overwrites.
- **\`manualSuggestion\` toggled mid-session.** Component re-reads from context; \`SuggestionPopover\` swaps branches; pending state resets on remount.
- **Multiple rapid clicks.** Button is disabled while pending — duplicate requests blocked at the UI layer.
- **Suggestion arrives in manual mode without a preceding click.** Auto-accepted defensively.
- **Telemetry in manual mode.** Keystroke \`track({ changed })\` is suppressed; only the manual-button click fires \`track({ requested })\`. \`blurred\`, \`focused\`, etc. still fire as today.

## Out of scope

- Manual mode for \`Select\`. (\`Checkbox\` and \`Switch\` also render \`SuggestionPopover\` and would inherit manual rendering, but \`InputControl\`'s debounce-skip change is Input/Textarea-specific.)
- Timeout, retry, or error UI for stuck pending state.
- A separate \`requestSuggestion\` API. We reuse the existing \`track\`.

## Test plan

- [ ] Storybook → \`Input\` → \`WithManualSuggestion\`: type, click Opal, verify spinner → auto-accept → button stays visible; click again to request another suggestion.
- [ ] Storybook → \`Textarea\` → \`WithManualSuggestion\`: same flow.
- [ ] Storybook → \`Input\` → \`WithSuggestion\` (auto mode): unchanged behavior, popover still appears.
- [ ] Lint, typecheck, vitest pass (\`pnpm lint\`, \`pnpm test\`).

### DIFF
--- a/.changeset/manual-suggestion-trigger.md
+++ b/.changeset/manual-suggestion-trigger.md
@@ -3,4 +3,4 @@
 "@optiaxiom/react": patch
 ---
 
-Add `manualSuggestion` flag to `unstable_SurfaceProvider`. When set, `Input` and `Textarea` skip the keystroke-debounced `track({ changed })` and instead render an Opal trigger button (pulses while pending) that fires `track({ changed, value })` on click. The next value-suggestion that arrives is auto-accepted silently. After accept the button hides until the value diverges again.
+Add `manualSuggestion` flag to `unstable_SurfaceProvider`. When set, `Input` and `Textarea` skip the keystroke-debounced `track({ changed })` and instead render an Opal trigger button that fires `track({ requested, value })` on click and shows a loading spinner while pending. The next value-suggestion that arrives is auto-accepted silently.
--- a/.changeset/manual-suggestion-trigger.md
+++ b/.changeset/manual-suggestion-trigger.md
@@ -1,0 +1,6 @@
+---
+"@optiaxiom/globals": patch
+"@optiaxiom/react": patch
+---
+
+Add `manualSuggestion` flag to `unstable_SurfaceProvider`. When set, `Input` and `Textarea` skip the keystroke-debounced `track({ changed })` and instead render an Opal trigger button (pulses while pending) that fires `track({ changed, value })` on click. The next value-suggestion that arrives is auto-accepted silently. After accept the button hides until the value diverges again.
--- a/packages/react/src/suggestion/SuggestionManualTrigger.tsx
+++ b/packages/react/src/suggestion/SuggestionManualTrigger.tsx
@@ -24,7 +24,6 @@ export const SuggestionManualTrigger = forwardRef<
   }, []);
 
   const [pending, setPending] = useState(false);
-  const [acceptedValue, setAcceptedValue] = useState<unknown>(undefined);
   const suggestions = useSuggestions("property", "value");
   const suggestion = suggestions?.find((s) => s.value !== surface?.value);
 
@@ -32,7 +31,6 @@ export const SuggestionManualTrigger = forwardRef<
     if (!suggestion) {
       return;
     }
-    setAcceptedValue(suggestion.value);
     surface?.accept(suggestion.id);
     setPending(false);
   });
@@ -40,7 +38,7 @@ export const SuggestionManualTrigger = forwardRef<
     acceptSuggestion();
   }, [suggestion?.id]);
 
-  if (!surface || surface.value === acceptedValue) {
+  if (!surface) {
     return null;
   }
 
@@ -58,7 +56,7 @@ export const SuggestionManualTrigger = forwardRef<
       }
       onClick={() => {
         setPending(true);
-        surface.track({ name: "changed", value: surface.value });
+        surface.track({ name: "requested", value: surface.value });
       }}
       ref={ref}
       size="sm"
--- a/packages/react/src/suggestion/SuggestionManualTrigger.tsx
+++ b/packages/react/src/suggestion/SuggestionManualTrigger.tsx
@@ -25,7 +25,7 @@ export const SuggestionManualTrigger = forwardRef<
 
   const [pending, setPending] = useState(false);
   const suggestions = useSuggestions("property", "value");
-  const suggestion = suggestions?.find((s) => s.value !== surface?.value);
+  const suggestion = suggestions?.[0];
 
   const acceptSuggestion = useEffectEvent(() => {
     if (!suggestion) {
@@ -47,13 +47,8 @@ export const SuggestionManualTrigger = forwardRef<
       appearance="subtle"
       aria-label="Opal suggestion"
       disabled={pending}
-      icon={
-        <Avatar
-          animation={pending ? "pulse" : undefined}
-          fallback="opal"
-          size="2xs"
-        />
-      }
+      icon={<Avatar fallback="opal" size="2xs" />}
+      loading={pending}
       onClick={() => {
         setPending(true);
         surface.track({ name: "requested", value: surface.value });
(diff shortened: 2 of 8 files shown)


packages/react/src/suggestion/SuggestionManualTrigger.tsx
```diff
@@ -0,0 +1,71 @@
...
+    surface?.accept(suggestion.id);
+    setPending(false);
+  });
+  useEffect(() => {
+    acceptSuggestion();
+  }, [suggestion?.id]);
+
+  if (!surface || surface.value === acceptedValue) {
+    return null;
+  }
```
radnan: let's not do this - only hide it if no surface - but always render it otherwise. user might want a new suggestion even without changing the existing value
```suggestion
  if (!surface) {
    return null;
  }
```

packages/react/src/suggestion/SuggestionManualTrigger.tsx
```diff
@@ -0,0 +1,71 @@
...
+      icon={
+        <Avatar
+          animation={pending ? "pulse" : undefined}
+          fallback="opal"
+          size="2xs"
+        />
+      }
+      onClick={() => {
+        setPending(true);
+        surface.track({ name: "changed", value: surface.value });
```
radnan: can we introduce a new track type instead of re-using `changed`

let's do `requested`

toha993: Addressed vai.

packages/react/src/suggestion/SuggestionManualTrigger.tsx
toha993: Addressed vai.

packages/react/src/suggestion/SuggestionManualTrigger.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const register = useEffectEvent(
+    surface?.suggestionPopover.register ?? (() => {}),
+  );
+  useEffect(() => {
+    return register();
+  }, []);
+
+  const [pending, setPending] = useState(false);
+  const suggestions = useSuggestions("property", "value");
+  const suggestion = suggestions?.find((s) => s.value !== surface?.value);
```
radnan: need to use the first suggestion instead of one that does not match surface.value

otherwise if the suggestion happens to match existing value then you get stuck in a perpetual "pending" state

packages/react/src/suggestion/SuggestionManualTrigger.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return (
+    <Button
+      appearance="subtle"
+      aria-label="Opal suggestion"
+      disabled={pending}
+      icon={
+        <Avatar
+          animation={pending ? "pulse" : undefined}
```
radnan: can we show a loading spinner instead of pulse animation. pulse is not very noticeable

.changeset/manual-suggestion-trigger.md
```diff
@@ -0,0 +1,6 @@
+---
+"@optiaxiom/globals": patch
+"@optiaxiom/react": patch
+---
+
+Add `manualSuggestion` flag to `unstable_SurfaceProvider`. When set, `Input` and `Textarea` skip the keystroke-debounced `track({ changed })` and instead render an Opal trigger button (pulses while pending) that fires `track({ changed, value })` on click. The next value-suggestion that arrives is auto-accepted silently. After accept the button hides until the value diverges again.
```
radnan: need to update this changelog to reflect actual code

radnan (approved):

PR #1578, opened 2026-05-18
fix(proteus): toast skipped-file count when selection exceeds maxFiles

vk1cd314:
When the user selects (or drops) more files than `maxFiles` allows, the SDK now uploads the files that fit and surfaces a warning toast via `@optiaxiom/react`'s `toaster` naming how many were skipped (e.g. *"Added 5 files. 2 were skipped because the limit is 10."*), instead of silently discarding the overflow.

No API change — host apps already mount the toaster provider, so the message renders without any wiring.

### DIFF
--- a/.changeset/sweet-dodos-float.md
+++ b/.changeset/sweet-dodos-float.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/proteus": patch
+---
+
+`ProteusFileUpload`: surface a warning toast naming the number of dropped files when the selection would exceed `maxFiles`, instead of silently discarding them. The allowed files still upload as before.
--- a/packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
+++ b/packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
@@ -1,5 +1,5 @@
 import { IconPlus } from "@optiaxiom/icons";
-import { Flex } from "@optiaxiom/react";
+import { Flex, toaster } from "@optiaxiom/react";
 import {
   FileUpload,
   FileUploadDropzone,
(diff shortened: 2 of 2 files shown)


packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
```diff
@@ -86,13 +99,17 @@ export function ProteusFileUpload({
...
-        incoming = incoming.slice(0, 1);
+      if (maxFiles !== undefined) {
+        const existing = multiple ? itemsRef.current.length : 0;
+        const remaining = Math.max(0, maxFiles - existing);
+        if (incoming.length > remaining) {
+          onOverflow?.({
+            maxFiles,
+            remaining,
+            selected: incoming.length,
+          });
```
tanvirstreame:
```suggestion
          onOverflow?.({
              limit: maxFiles,
              available: remaining,
              attempted: incoming.length,
          });
```

packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
```diff
@@ -86,13 +86,16 @@ export function ProteusFileUpload({
...
+      if (maxFiles !== undefined) {
+        const existing = multiple ? itemsRef.current.length : 0;
+        const remaining = Math.max(0, maxFiles - existing);
+        if (incoming.length > remaining) {
+          toaster.create(
+            `Too many files selected. You can add up to ${remaining} more file${remaining !== 1 ? "s" : ""} (maximum ${maxFiles}).`,
+            { type: "danger" },
+          );
+          return;
+        }
```
radnan, line 96: let's not reject all files here

instead we should accept however many files are allowed - and also show a toast alerting the user and noting the number of files that were dropped.

"Added 5 files. 2 were skipped because the limit is 10."

.changeset/sweet-dodos-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/proteus": patch
+---
+
+adds onOverflow to handle cases when more than max number of files are added
```
radnan: update the patch message here and commit message and PR title/description all to match the impl

vk1cd314: addressed

packages/proteus/src/proteus-file-upload/ProteusFileUpload.tsx
vk1cd314: addressed

radnan (approved):

PR #1608, opened 2026-06-02
Add rich text editor component

tanvirstreame:
## Summary
- Adds `RichTextEditor` to `@optiaxiom/react` — a Tiptap-based WYSIWYG editor with a toolbar for bold, italic, bullet/numbered lists, and links
- Toolbar icons use `@optiaxiom/icons` (new workspace dep on `@optiaxiom/react`)
- Exposes `ax-rich-text-editor` as a custom element via `@optiaxiom/web-components`
- Supports `defaultValue` (uncontrolled), `value` + `onValueChange` (controlled), `readOnly`, and `placeholder`

## Test plan
- [ ] Storybook: `RichTextEditor > Basic / DefaultValue / Controlled / Readonly` render and behave as expected
- [ ] Toolbar buttons toggle the corresponding marks in the editor
- [ ] Read-only mode hides the toolbar
- [ ] Controlled mode reflects parent state and calls `onValueChange` on edits
- [ ] Vitest suite passes (`pnpm test`) — 4 RichTextEditor specs


packages/react/src/rich-text-editor/RichTextEditor.tsx
```diff
@@ -0,0 +1,191 @@
...
+     */
+    value?: string;
+  }
+>;
+
+/**
+ * Rich text editor with a toolbar for basic formatting (bold, italic, lists, links).
+ *
+ * @category form
+ * @since 1.10.0
```
radnan, line 55: also add `@experimental` here

packages/react/src/rich-text-editor/RichTextEditor.tsx
```diff
@@ -0,0 +1,191 @@
...
+     */
+    placeholder?: string;
+    /**
+     * Whether the editor is read-only.
+     */
+    readOnly?: boolean;
+    /**
+     * Whether the field is required for form submission.
+     */
+    required?: boolean;
```
radnan: can we actually enforce this? if not let's remove it

packages/react/package.json
```diff
@@ -77,6 +78,11 @@
     "@radix-ui/react-visually-hidden": "^1.2.3",
     "@tanstack/react-virtual": "^3.13.23",
     "@tanstack/virtual-core": "^3.13.23",
+    "@tiptap/extension-link": "^2.10.3",
+    "@tiptap/extension-placeholder": "^2.10.3",
+    "@tiptap/pm": "^2.10.3",
+    "@tiptap/react": "^2.10.3",
+    "@tiptap/starter-kit": "^2.10.3",
```
radnan: let's add these to peer dependencies instead

radnan (approved):

packages/react/src/rich-text-editor/RichTextEditor.tsx
tanvirstreame: Addressed

packages/react/src/rich-text-editor/RichTextEditor.tsx
tanvirstreame: Addressed by removing

packages/react/package.json
tanvirstreame: Addressed

PR #1649, opened 2026-06-11
Fix rich text editor scrollable height

tanvirstreame:
## Summary
Screenshot 2026-06-11 at 9 25 33 PM

When `RichTextEditor` is rendered with a constrained height (e.g. `<RichTextEditor h="224" />`), the editor content currently overflows the container — the toolbar scrolls along with the content, and the visual boundary is broken.

This patch makes the editor handle constrained heights correctly: the toolbar stays pinned at the top while the editable content scrolls inside its own area.

## Changes

**`packages/react/src/rich-text-editor/RichTextEditor.css.ts`** — the actual fix
- Outer editor container is now a flex column with `overflow: hidden`
- The ProseMirror wrapper (`> div:has(> .ProseMirror)`) gets `flex: 1`, `min-height: 0`, `overflow-y: auto` — so it claims the remaining vertical space and scrolls on its own

**`apps/storybook/src/components/RichTextEditor.stories.tsx`** — `FixedHeight` story that demonstrates the behavior with `h="224"` and enough paragraphs to overflow.

**Patch bump for `@optiaxiom/react`** via changeset.

## Before / after

| Before | After |
|---|---|
| Toolbar scrolls with content; content can overflow the rounded border | Toolbar stays pinned; content scrolls inside the bordered area |

## Test plan

- [ ] Run `pnpm --filter @optiaxiom/storybook dev`, navigate to **Components → RichTextEditor → Fixed Height** — verify toolbar is pinned and the inner content area has its own scrollbar
- [ ] Other RichTextEditor stories (Basic, Default Value, Controlled, Readonly) render unchanged — no height set, content sizes naturally
- [ ] Lint and existing tests still pass: `pnpm lint --no-fix && pnpm test`

### DIFF
--- a/packages/react/src/rich-text-editor/RichTextEditor.css.ts
+++ b/packages/react/src/rich-text-editor/RichTextEditor.css.ts
@@ -9,6 +9,8 @@ export const editor = recipe({
     {
       bg: "bg.default",
       border: "1",
+      display: "flex",
+      flexDirection: "column",
       rounded: "md",
     },
     style({
@@ -39,6 +41,12 @@ export const toolbar = recipe({
       flexWrap: "wrap",
     }),
   ],
+});
+
+globalStyle(`${editorClass} > div:has(> .ProseMirror)`, {
+  flex: 1,
+  minHeight: 0,
+  overflowY: "auto",
 });
 
 globalStyle(`${editorClass} .ProseMirror`, {
--- a/.changeset/rich-text-editor-scrollable-height.md
+++ b/.changeset/rich-text-editor-scrollable-height.md
@@ -1,0 +1,8 @@
+---
+"@optiaxiom/react": patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/react/src/rich-text-editor/RichTextEditor.css.ts
```diff
@@ -9,6 +9,8 @@ export const editor = recipe({
     {
```
radnan, line 9: hmm i'm noticing some issues with spacing in component (unrelated to this change) - let me give a PR for that before you do a release

tanvirstreame: Okay

radnan:
Heads up: #1655 is stacked on top of this branch (RichTextEditor content styling + table support). It should merge after this one.

vk1cd314 (approved):

PR #1660, opened 2026-06-15
feat: Card hover state + document file-icon glyph

vk1cd314:
## What

Design-system support for the Opal file-card redesign (OPAL-4092).

- **Card** (`packages/react/src/card/Card.css.ts`): add a hover border state to the card recipe (uses the `border.secondary` token, `(hover: hover)` only) so cards show an interactive affordance on hover.
- **ProteusFileIcon** (`packages/proteus/src/proteus-file-icon/ProteusFileIcon.tsx`): map text / pdf / word document mime types to the folded-corner document glyph (`IconFileWord`) to match the file-card icon directory in the design.
- **Storybook** (`apps/storybook/src/components/ProteusDocumentRenderer.stories.tsx`): update the `WithFiles` story to the redesigned file card (preview button + the file-type icon set) and add a `WithFilesSelected` story so both the hover and selected states are reviewable.

## Why

The Opal file/artifact action cards are being reworked to show a hover state and a file-type icon set that matches the design. Both are owned by the design system, so they live here; the consuming template changes live in opal-app (newscred/opal-app#5160).

## Notes

- No API or schema changes — hover is pure CSS; the icon change swaps one glyph in the existing mime→icon map.

## Screenshots
image


packages/react/src/card/Card.css.ts
```diff
@@ -26,6 +26,16 @@ export const card = recipe({
 
       borderRadius: theme.borderRadius.lg,
       position: "relative",
+
+      "@media": {
```
radnan: let's not change Card - rather create a ProteusCard component and make the changes there

also this should not have a bare hover effect - rather it should only show hover if hovering over a CardLink inside the Card

PR #1683, opened 2026-06-22
fix(react): make primary-opal render identically to primary

toha993:
## Summary

Makes the `primary-opal` button appearance render identically to `primary` for backward compatibility. Existing consumers of `appearance="primary-opal"` keep working but now get the plain (non-animated) primary look.

## Changes
- `primary-opal` now maps to `{ intent: "primary", variant: "strong" }` — same as `primary`.
- Dropped the now-unreachable `strong-opal` branches (spinner color + `staticVariant`) since nothing produces that variant anymore. This also makes the loading spinner match `primary` exactly.
- Added a patch changeset for `@optiaxiom/react`.

## Notes
- Did not delete the `strong-opal` CSS variant itself, in case an external consumer still references it.

### DIFF
--- a/.changeset/primary-opal-backward-compat.md
+++ b/.changeset/primary-opal-backward-compat.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/react": patch
+---
+
+Make the `primary-opal` button appearance render identically to `primary` for backward compatibility.
--- a/packages/react/src/button/ButtonRoot.css.ts
+++ b/packages/react/src/button/ButtonRoot.css.ts
@@ -282,18 +282,6 @@ export const buttonBase = recipe({
           },
         },
       }),
-      "strong-opal": style({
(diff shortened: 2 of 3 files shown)


.changeset/primary-opal-backward-compat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/react": patch
+---
+
+Make the `primary-opal` button appearance render identically to `primary` for backward compatibility.
```
radnan, line 5: let's also delete the strong-opal variant from the css entirely

toha993: Done — deleted the `strong-opal` variant style and its sm compound-variant from `ButtonRoot.css.ts` (no references remain; button tests pass).

radnan (approved):

PR #1684, opened 2026-06-22
feat(react): update inlined opal icon artwork

toha993:
## Summary

Updates the inlined `IconOpal` image to the current Opal orb artwork. Because `IconOpal` is the single source for the opal logo, this updates **both** consumers at once:
- `Alert` (opal intent icon)
- `Avatar` with `fallback="opal"`

image

## Approach
Kept the existing inline-base64 pattern (the whole library inlines images — there are no asset-file imports, and it builds with rolldown). The new orb was re-encoded to a **96px WebP (~2.3KB)** with `cwebp` — same order of magnitude as the previous 48px asset (1.6KB), just retina-sharp. No CDN dependency, no bundle bloat.

## Changes
- `IconOpal.tsx`: swapped the base64 `src` for the new orb.
- Patch changeset for `@optiaxiom/react`.


packages/react/src/avatar/Avatar.tsx
```diff
@@ -3,10 +3,12 @@ import * as RadixAvatar from "@radix-ui/react-avatar";
 import { forwardRef } from "react";
 
 import { Box, type BoxProps } from "../box";
-import { IconOpal } from "../icons/IconOpal";
 import * as styles from "./Avatar.css";
 import { useAvatarContext } from "./AvatarContext";
 
+const OPAL_AVATAR_SRC =
```
radnan: just inline the png as like a base64 string

toha993: Done — switched away from the CDN URL. It's now inlined as a base64 string directly in `IconOpal` (re-encoded to a 96px WebP, ~2.3KB), so both the Alert opal icon and the `fallback="opal"` avatar update from one place. This PR no longer touches Avatar.tsx.

radnan (approved):

PR #1718, opened 2026-06-26
ci: flag missing @optiaxiom/mcp changesets vs last published release

radnan:
## Problem

`@optiaxiom/mcp`'s published `dist` is **generated** at build time from `@optiaxiom/react`, `@optiaxiom/globals`, and `@optiaxiom/shared` (via `packages/mcp/plugins/generators.mjs`). When someone changes a font token, a prop description, or adds an icon, they add a changeset for `react`/`globals` — but nothing tells them the **MCP output also changed**, so `mcp` never gets bumped and silently falls behind (confirmed: `main` already differs from published `3.0.2`).

`changeset-bot` only checks whether a changeset *file* exists — it can't know a `globals`/`react` edit *should* also bump `mcp`. This adds the missing signal.

## Approach

1. **Deterministic generation (prerequisite).** A content hash is only a valid drift signal if the data is byte-stable across machines. Three `readdir` sites ordered demo/test collection by on-disk order (macOS sorted, Linux inode-order). Sorted them. Prop/enum-value order (e.g. `padding: 2 | 4 | 6`) comes from the committed `docs.json` + react-docgen declaration order — already stable and **semantically meaningful**, so it's preserved (not sorted).

2. **`@optiaxiom/mcp/data` subpath export.** The package now exposes its raw generated `components`/`guides`/`icons`/`tests`/`tokens` (like `@optiaxiom/proteus/spec`), so the check imports the package's own data instead of a hardcoded list.

3. **Compare against the last published release, not the PR base branch.** `oas-changeset-drift` builds the local data, `npm pack`s the latest `@optiaxiom/mcp`, and hashes each section to diff. Comparing against the **release** is deliberate: if you compared against `main`, merging an unbumped change once would make it the new baseline and **permanently silence** the warning. Against the release, the drift keeps being flagged until a release actually ships the changeset.

4. **`comment.yml`** posts a changeset-bot-style ⚠️ comment when mcp output drifted since the last release with no changeset, and clears it once resolved. Non-blocking (`|| true`). The check is self-contained (no base checkout), so it's a single PR.

## Changesets

- `@optiaxiom/mcp` **minor** — new `./data` public entrypoint.
- `@optiaxiom/mcp` **patch** — regenerated data already stale on `main` (font tokens, fonts guide, new icons) + deterministic generation.

## Verification

- Determinism: `pnpm --silent changeset-drift hash`-equivalent digest is identical across fresh runs.
- `node -e "import('@optiaxiom/mcp/data').then(d=>console.log(Object.keys(d)))"` → `components, guides, icons, tests, tokens`.
- Drift states (local build vs published `3.0.2`): matches release → exit 0; changed + no changeset → ⚠️ + exit 1; changed + changeset → ✅ + exit 0.
- `tsgo -b`, `oxlint --type-aware`, `oxfmt`, `vitest --project=@optiaxiom/mcp` all pass. `npm pack` includes `dist/src/data.js` + `dist/data.d.ts`.

### DIFF
--- a/.changeset/mcp-data-export.md
+++ b/.changeset/mcp-data-export.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/mcp": minor
+---
+
+Added a `@optiaxiom/mcp/data` subpath export exposing the raw generated component, guide, icon, test, and token metadata for programmatic consumers.
--- a/.changeset/mcp-regenerate-data.md
+++ b/.changeset/mcp-regenerate-data.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/mcp": patch
+---
+
+Regenerated the bundled component, token, icon, and guide data so the MCP server reflects the latest `@optiaxiom/react` and `@optiaxiom/globals` output — including updated font family tokens, the new fonts guide, and newly added icons. Demo and test collection is now sorted so generated output is deterministic across platforms.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ jobs:
           echo ${{ github.event.number }} > ./pr/ID
           pnpm build
           pnpm --silent bundle-size compare bundle-size-base.json > ./pr/bundle-size.md
+          pnpm --silent changeset-drift > ./pr/changeset-drift.md || true
 
       - uses: actions/upload-artifact@v7
         with:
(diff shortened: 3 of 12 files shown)


radnan:
Split into #1720 (the `@optiaxiom/mcp/data` export, released first) and the stacked tooling PR that compares against the last npm release. Closing this in favor of that pair.

PR #1719, opened 2026-06-26
ci: run changeset-drift snapshot/compare in build job

radnan:
> **Stacked — part 2 of 2.** Depends on #1718. Base is the part-1 branch, so this diff is just the `ci.yml` wiring. Merge #1718 first, then retarget this to `main` (or merge after).

## What

Wires the `oas-changeset-drift` tool (added in #1718) into the existing `build` job:

- **Base checkout**: `pnpm --silent changeset-drift snapshot > changeset-drift-base.json`, uploaded alongside `bundle-size-base.json`.
- **PR checkout**: `pnpm --silent changeset-drift compare changeset-drift-base.json > ./pr/changeset-drift.md || true`, shipped in the existing `pr-build` artifact for `comment.yml` to surface.

The `compare` uses `|| true` so a missing changeset never fails the build — the signal is the PR comment, matching the `changeset-bot` non-blocking behavior.

## Why separate from #1718

The base checkout runs against `main`, so the `oas-changeset-drift` command must already exist there. #1718 lands the command on `main`; this PR can only safely merge afterward. (The first attempt that combined both failed here: the base checkout on `main` had no `changeset-drift` command and the build step exited 254.)

### DIFF
--- a/.changeset/mcp-regenerate-data.md
+++ b/.changeset/mcp-regenerate-data.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/mcp": patch
+---
+
+Regenerated the bundled component, token, icon, and guide data so the MCP server reflects the latest `@optiaxiom/react` and `@optiaxiom/globals` output — including updated font family tokens, the new fonts guide, and newly added icons.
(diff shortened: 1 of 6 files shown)


radnan:
Superseded by #1718 — the new approach is a single PR (hash-based drift check, no stacked ci.yml change needed).

PR #1721, opened 2026-06-26
ci: flag missing @optiaxiom/mcp changesets vs last published release

radnan:
> **Depends on #1720** (the `@optiaxiom/mcp/data` export). Merge + release #1720 first; until a release ships the export, this check skips. This PR is stacked on #1720, so its diff shows only the tooling once #1720 merges.

## Problem

`@optiaxiom/mcp`'s published `dist` is **generated** at build time from `@optiaxiom/react`, `@optiaxiom/globals`, and `@optiaxiom/shared`. Its consumer-facing output can change without anyone touching the package, so the changeset that bumps it gets forgotten (`main` already differs from published `3.0.2`). `changeset-bot` only checks whether a changeset *file* exists — it can't know a `globals`/`react` edit *should* bump `mcp`.

## Approach

`oas-changeset-drift`:
- imports the local `@optiaxiom/mcp/data` export and the **same export from the last npm release** (installed into a temp dir, resolved via the package's own `exports` map);
- hashes each section and flags drift that has no matching `@optiaxiom/mcp` changeset.

**Why compare against the release, not the PR base branch:** if you diffed against `main`, merging an unbumped change once would make it the new baseline and **permanently silence** the warning. Against the release, it keeps firing until a release actually ships the changeset.

**Sections are derived from the data** (`Object.keys`), so a newly added section is covered automatically — nothing hardcoded.

**Determinism (prerequisite):** three `readdir` sites ordered demo/test collection by on-disk order (macOS sorted, Linux inode-order). Sorted them so the hash is reliable across platforms. Prop/enum-value order is unchanged (it comes from `docs.json` + react-docgen and is semantically meaningful).

`ci.yml` writes the report in the build job; `comment.yml` posts a changeset-bot-style comment when output drifted with no changeset, and clears it once resolved. Non-blocking.

## Changeset

- `@optiaxiom/mcp` **patch** — deterministic data generation (sorted demo/test collection).

## Verification

- Against real npm (`3.0.2`, no `./data` yet) → skips cleanly, exit 0.
- Against a simulated release that has `./data`: identical → no drift; mutated local section → reports exactly that section as changed.
- `tsgo -b`, `oxlint --type-aware`, `oxfmt` clean.

### DIFF
--- a/.changeset/mcp-data-export.md
+++ b/.changeset/mcp-data-export.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/mcp": minor
+---
+
+Added a `@optiaxiom/mcp/data` subpath export exposing the raw generated component, guide, icon, test, and token metadata for programmatic consumers.
--- a/.changeset/mcp-deterministic-data.md
+++ b/.changeset/mcp-deterministic-data.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/mcp": patch
+---
+
+Sorted demo and test collection during data generation so the bundled metadata is deterministic across platforms.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ jobs:
           echo ${{ github.event.number }} > ./pr/ID
           pnpm build
           pnpm --silent bundle-size compare bundle-size-base.json > ./pr/bundle-size.md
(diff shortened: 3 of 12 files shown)


radnan:
Superseded: switching to a committed src/data.json gated by git diff (proteus-style) — a hard check that can't be ignored, no tooling/export needed.

PR #1725, opened 2026-06-28
Add proteus outlook icon

tanvirstreame:
## Summary

- Adds `IconOutlook` to `@optiaxiom/icons` using the official Microsoft Outlook (2025) multi-color brand SVG.
- Teaches `materialSymbolsPlugin` to handle multi-path / multi-color SVGs alongside the existing single-path `MaterialIcon` pipeline.

## Why

Opal-app needs an `Outlook` entry in its `PROTEUS_ICONS` map (used by `{ $type: "Icon", name: "Outlook" }` in Proteus documents). Existing `@optiaxiom/icons` icons are single-path monochrome glyphs from Material Symbols — none match the Outlook brand visually. This adds the brand glyph as a first-class import: `import { IconOutlook } from "@optiaxiom/icons"`.

## How it works

- The Outlook SVG is hand-committed to [`packages/icons/svg/outlook.svg`](packages/icons/svg/outlook.svg) (force-added past the `svg/*.svg` gitignore that auto-fetched Material SVGs live behind). A mirror `outlook-fill.svg` exists so `download.mjs`'s `existsSync` skip prevents a futile Google Fonts fetch for the fill variant.
- The rolldown plugin (`materialSymbolsPlugin`) now branches on SVG content: if the file contains `<defs>` or more than one `<path>`, it emits a `dangerouslySetInnerHTML`-backed `forwardRef` that preserves the original viewBox + inline fills/gradients. Single-path Material Symbols icons go through the existing `MaterialIcon` wrapper unchanged.
- The brand-icon code applies sizing (`flex: none`, `width`/`height`) inline via the `style` prop instead of importing `MaterialIcon.css` — the generated chunk lands inside a deeply nested `_virtual/` path where the relative CSS import would 404.

## Test plan

- [ ] `pnpm -F @optiaxiom/icons build` succeeds and emits a `_virtual/.../outlook.js` chunk
- [ ] `pnpm -F @optiaxiom/storybook dev` → `CreateMeetingEvent` story → click **Create event** → the Outlook glyph renders in the success view alongside the Clock / LocationDot rows
- [ ] No console 404s when loading the story
- [ ] `IconOutlook` shows up in `dist/index.d.ts`
- [ ] Existing icons (e.g. `IconUserGroup`, `IconClock`) still render unchanged in their stories

### DIFF
--- a/.changeset/mcp-deterministic-data.md
+++ b/.changeset/mcp-deterministic-data.md
@@ -1,5 +1,0 @@
----
-"@optiaxiom/mcp": patch
----
-
-Regenerated the bundled metadata and made its generation deterministic so the committed data stays in sync with `@optiaxiom/react` and `@optiaxiom/globals`. The `get_patterns` example `code` is now a list of `{ filename, content }` objects (in display order) instead of a filename-to-contents map.
--- a/packages/mcp/CHANGELOG.md
+++ b/packages/mcp/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @optiaxiom/mcp
 
+## 3.0.3
+
+### Patch Changes
+
+- 068edf7: Regenerated the bundled metadata and made its generation deterministic so the committed data stays in sync with `@optiaxiom/react` and `@optiaxiom/globals`. The `get_patterns` example `code` is now a list of `{ filename, content }` objects (in display order) instead of a filename-to-contents map.
+
 ## 3.0.2
 
 ### Patch Changes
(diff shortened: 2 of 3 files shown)


packages/icons/icons.json
```diff
@@ -203,6 +203,7 @@
   "open_in_full": ["ArrowsMaximize"],
   "open_in_new": ["ArrowUpRightFromSquare"],
   "orders": ["BoxOpen"],
+  "outlook": ["Outlook"],
```
radnan: We don't need to add this icon here - just add it directly to ProteusMessageCard

PR #1750, opened 2026-08-03
feat(proteus): add requestSidebar action variant + onRequestSidebar handler

kamrul1157024:
Mirrors `requestModal` (d746979) file for file, and adds an optional `auto` declaration on the same handler.

## `requestSidebar`

Same payload as `requestModal` — only the surface differs. A modal covers the thread; a sidebar sits beside it, so the user can keep reading while a fuller view stays open. Prefer it when the panel is something to work *alongside* rather than a detour.

```jsonc
{
  "$type": "Action",
  "children": "Open results",
  "onClick": {
    "action": "requestSidebar",
    "resource": "ui://experiment-results",
    "title": "Experiment results",
    "params": { "experiment": { "$type": "Value", "path": "/experiment" } }
  }
}
```

`ProteusDocumentShell` gains `onRequestSidebar`; a host that doesn't implement the panel omits the prop and the action is a no-op. `params` needed no special handling — `ProteusAction` already runs the whole `onClick` through `useResolveProteusValues`, which recurses into plain objects (same reason the `requestModal` commit didn't touch it).

## `auto: { when }`

Declares that a host **may** fire the handler without a click, using the same condition grammar as `Show`.

Putting it on the handler rather than on the `Document` is deliberate: a card that auto-opens then always has a **visible control to reopen the panel** after the user closes it. A document-level property makes that dead end easy to ship by accident. The cost — `onClick` now names something that can fire without a click — is accepted for that guarantee.

### Why the renderer has to do the work

Hosts can't evaluate this themselves: `when` and `params` contain ProteusExpressions and the engine lives in this package. So `useAutoSidebar`:

- walks the document for a handler carrying `auto` — **first in document order wins**, and the losers keep working as ordinary click handlers;
- evaluates the condition and resolves `params`;
- hands the host a plain payload flagged **`auto: true`**, so a host can gate unprompted opens without gating clicks.

**Edge-triggered on `when` going false → true**, not fire-once-on-mount — matching `watch()` semantics. Fire-once can't express *"open the panel when the job finishes"*, which is the main use case.

`auto` itself is **not** forwarded to the callback: it describes when a host may fire the handler, which is decided *before* dispatch.

The hook runs before `ProteusDocumentRenderer`'s early return so hook order stays stable across valid and invalid documents.

## Checks

- `pnpm -F @optiaxiom/proteus build` — both `schema/*.json` regenerated by the generator plugin (not hand-edited)
- `pnpm lint` clean
- Changeset included (`minor`)
- Storybook story + a "Sidebar handler" section in the Proteus guide, matching the `requestModal` pair

Validated against the regenerated runtime and public schemas: a `requestSidebar` document with `auto`, `params` and `title` passes; the new prop type-checks from a consuming app.

## Unrelated pre-existing issue, noticed while testing

The `onClick` `anyOf` accepts almost anything — a `requestModal` missing its required `resource`, an unknown `action`, even `{}`. A permissive member swallows everything, so malformed handlers inside a document literal are caught by neither the JSON schema nor TypeScript (the `element` prop's nested types are loose); only renderer props are genuinely type-checked.

Combined with `ProteusDocumentRenderer` returning `null` *silently* on a document that fails validation, an authoring mistake presents as an empty render with no console error. Not touched here — happy to open a separate issue if useful.

### DIFF
--- a/.changeset/proteus-request-sidebar-action.md
+++ b/.changeset/proteus-request-sidebar-action.md
@@ -1,0 +1,5 @@
+---
+"@optiaxiom/proteus": minor
+---
+
+Added a `requestSidebar` client-side `Action.onClick` variant and matching `onRequestSidebar({ resource, params, title })` handler on `ProteusDocumentShell` / `ProteusDocumentRenderer` so hosts can dock a named `ui://...` resource in a side panel alongside the conversation instead of over it. The variant also accepts an optional `auto: { when }` describing when a host may fire the handler without a click — hosts that do not support it ignore the field.
--- a/apps/docs/app/(docs)/guides/proteus/page.mdx
+++ b/apps/docs/app/(docs)/guides/proteus/page.mdx
@@ -168,6 +168,43 @@ Opens a modal on the host that renders a named UI resource of the same tool prov
 
 The host wires this via the `onRequestModal({ resource, params })` prop on `ProteusDocumentRenderer` — the library is agnostic about how the modal itself is rendered. Federated remotes can also fire the event by calling `onEvent({ action: "requestModal", ... })` directly.
 
+### Sidebar handler
+
+`requestSidebar` carries the same payload as `requestModal` — only the surface differs. Where a modal covers the conversation, a sidebar docks beside it, so the user can keep reading the thread while a fuller view stays open. Prefer it when the panel is something to work alongside rather than a detour. An optional `title` sets the panel header, falling back to the target document's own `title`.
+
+```json
+{
+  "$type": "Action",
+  "appearance": "primary",
+  "children": "Open results",
+  "onClick": {
+    "action": "requestSidebar",
+    "resource": "ui://experiment-results",
+    "title": "Experiment results",
+    "params": {
+      "experiment": { "$type": "Value", "path": "/experiment" }
+    }
+  }
+}
+```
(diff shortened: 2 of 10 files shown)


.changeset/proteus-request-sidebar-action.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/proteus": minor
+---
+
+Added a `requestSidebar` client-side `Action.onClick` variant and matching `onRequestSidebar({ resource, params, title })` handler on `ProteusDocumentShell` / `ProteusDocumentRenderer` so hosts can dock a named `ui://...` resource in a side panel alongside the conversation instead of over it. The variant also accepts an optional `auto: { when }` describing when a host may fire the handler without a click — hosts that do not support it ignore the field.
```
radnan, line 5: I'm not sure a separate event is needed for this

We use requestModal to allow opening a new resource window - but don't dictate whether it should be an actual modal or a sidebar

For now in Opal requestModal is implemented to be exactly like canvas and opens in an expanded sidebar while the thread is still open on the side. So this prop wouldn't be needed.

If we do want other options it'd be better to add on to requestModal rather than adding an entirely new event - to prevent confusion on the user end as to pick which one.

PR #1751, opened 2026-08-04
feat: add 800 (ExtraBold) to fontWeight scale

mauricecruz:
## Summary

Adds `800` (ExtraBold) to the `@optiaxiom/react` `fontWeight` scale.

## The gap

The `fontWeight` prop is a closed enum that jumps straight from `700` to `900`. In the shipped types (`packages/react/dist/css-runtime.d.ts`):

```ts
fontWeight?: "400" | "500" | "600" | "700" | "900" | "inherit" | undefined;
```

There is no `800`. However, the `heading` font family (VC Nudge) ships an ExtraBold (800) instance, so the weight exists in the font but could not be selected through the prop.

## What changed

- `packages/react/src/sprinkles/properties.css.ts` — added `"800"` to the `fontWeight` sprinkles scale, in sorted position between `"700"` and `"900"`. This single Vanilla Extract array is the source for both the prop type and the runtime atom, so the generated types and the emitted CSS class both gain `800`.
- `.changeset/fontweight-800-extrabold.md` — `minor` changeset for `@optiaxiom/react`.

## Verification

`pnpm -F "@optiaxiom/react..." build` succeeds. The regenerated (gitignored) type output now reads:

```ts
fontWeight?: "400" | "500" | "600" | "700" | "800" | "900" | "inherit" | undefined;
```

in both `dist/css-runtime.d.ts` and `dist/index-*.d.ts`. No `dist/` output is committed.

cc @radnan

_(Tracked internally at Optimizely as OPAL-5328.)_

### DIFF
--- a/packages/react/src/sprinkles/properties.css.ts
+++ b/packages/react/src/sprinkles/properties.css.ts
@@ -191,7 +191,7 @@ export const unresponsiveProps = defineProperties({
     /**
      * Set the element's `font-weight` CSS property
      */
-    fontWeight: ["400", "500", "600", "700", "900", "inherit"] as const,
+    fontWeight: ["400", "500", "600", "700", "800", "900", "inherit"] as const,
     marginBottom: margins,
     marginLeft: margins,
     marginRight: margins,
--- a/.changeset/fontweight-800-extrabold.md
+++ b/.changeset/fontweight-800-extrabold.md
@@ -1,0 +1,6 @@
+---
+"@optiaxiom/react": minor
+"@optiaxiom/mcp": patch
+---
+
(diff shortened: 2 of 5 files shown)


packages/react/src/sprinkles/properties.css.ts
```diff
@@ -191,7 +191,7 @@ export const unresponsiveProps = defineProperties({
     /**
      * Set the element's `font-weight` CSS property
      */
-    fontWeight: ["400", "500", "600", "700", "900", "inherit"] as const,
+    fontWeight: ["400", "500", "600", "700", "800", "900", "inherit"] as const,
```
radnan, line 194: the font file itself only ships 800 and when using 900 css will shift to the next closest weight which is 800- so this way we pick a good font weight for the web fallback font as well as the actual nudge font. that's why this discrepancy exists. closing the PR for now.

PR #1756, opened 2026-08-07
fix(react): align Heading level=2 with Figma Header/Heading 2 style

tanvirstreame:
## Summary

Aligns `<Heading level="2">` with the published Figma `Header/Heading 2` style. The token and the design library had drifted:

- `line-height`: 40px (code) vs 1.04 × 28 ≈ 29px (Figma) → tightened code default to **30px** (closest 2px-grid value to 29.12), written as `1.875rem` so it scales with the user's root font-size.
- `font-weight`: 700 (code) vs 800 (Figma) → bumped code default to **800**.
- Added `"800"` to the accepted `fontWeight` sprinkle scale (previously jumped `"700"` → `"900"`) so the new default is expressible.
- Regenerated `packages/proteus/src/schema/*.json` and `packages/mcp/src/data.json` to propagate the widened enum.

Also picks up a pre-existing regen gap — the `bg.message.neutral` color token added in `e8fcf32c` was never propagated to `packages/mcp/src/data.json`. CI on this branch flagged the drift; regen commit fills it in.

Discovered while implementing the Virtual Teammate resume card title; the consumer app currently carries an interim unitless `line-height: 1.04` override that can be removed once this ships.

## Behavior matrix for `<Heading level="2">`

Both new defaults (LH + weight) are scoped to the **default `fontSize`** case. If consumers override `fontSize` to something non-`3xl`, they fall back to the previous behavior — no silent regression on smaller h2s.

| Usage                                                     | line-height             | font-weight            |
|-----------------------------------------------------------|-------------------------|------------------------|
| `<Heading level="2">` (defaults)                          | **1.875rem** (30px, was 40px) | **800** (was 700)      |
| `<Heading level="2" fontSize="3xl">` (explicit default)   | **1.875rem** (30px)     | **800**                |
| `<Heading level="2" fontSize="lg">` (or any non-3xl)      | token-derived (24px)    | **700** (unchanged)    |
| `<Heading level="2" fontWeight="600" ...>` (explicit)     | per fontSize rule above | **600** (consumer wins)|

Concrete real-world check: [`packages/proteus/src/proteus-document/ProteusDocumentShell.tsx:431`](https://github.com/optimizely-axiom/optiaxiom/blob/main/packages/proteus/src/proteus-document/ProteusDocumentShell.tsx#L431) uses `<Heading fontSize="lg" fontWeight="600" level="2" ...>` — verified to render exactly as before (16px, 24px LH, weight 600).

## Scope

Only `<Heading level="2">` is affected. Other heading levels, `<Text fontSize="3xl">`, and every other `3xl`-token consumer are unchanged. Letter-spacing already matched (`tracking="wider"` resolves to 0.56px at 28px, matches Figma).

The `3xl` token in `@optiaxiom/globals` was intentionally NOT touched — the fix lives at the component layer so the token scale stays on its consistent 4px grid.

## Commits

Split into bisect-safe commits:

1. `feat(react): allow fontWeight="800" in sprinkle scale` — self-contained additive change.
2. `fix(react): align Heading level=2 with Figma Header/Heading 2 style` — the recipe + component + changeset.
3. `chore(proteus,mcp): regenerate schemas + data for fontWeight="800"` — generated artifacts for the sprinkle change.
4. `chore(mcp): regenerate data.json to include bg.message.neutral token` — pre-existing drift picked up by CI.
5. `fix(react): scope Heading level=2 line-height override to default fontSize` — Copilot-flagged guard so the 30px LH doesn't leak onto overridden fontSizes.
6. `fix(react): scope Heading level=2 fontWeight bump to default fontSize` — parallel guard for the same class of bug on font-weight.

## Semver

`minor` for `@optiaxiom/react`, `@optiaxiom/web-components`, `@optiaxiom/proteus`, `@optiaxiom/mcp` (changeset included).

## Heads-up for consumers

`<Heading level="2">` with the default `fontSize` will render **tighter (30px vs 40px LH)** and **heavier (800 vs 700)** after upgrading. Consumer apps with visual-regression / snapshot tests hitting default h2s should expect diffs. Downstream layout that relied on the previous 40px vertical rhythm around h2s may reflow slightly. Consumers that override `fontSize` on level 2 headings are unaffected.

**Opt-out semantics** (for case-by-case escape):
- Passing an explicit **`fontSize`** fully opts out — restores both the `700` weight and the token's `40px` line-height.
- Passing only an explicit **`fontWeight`** overrides the weight, but the tighter `30px` line-height still applies. If both need to be restored, pass `fontSize` too.

## Design sign-off

This PR provisionally lands the Figma-matching values in Axiom. If instead the Figma library should be updated to match the previous `3xl` token (40px LH, weight 700, keeping the whole type scale on a 4px grid), we can revert the source changes and open a Figma library ticket instead — the change is scoped enough that the revert would be trivial.

## Test plan

- [x] `pnpm lint` passes.
- [x] `pnpm test` passes (48/48).
- [ ] Storybook: `<Heading level="2">` in the `Levels` and `Sizes` stories renders tighter + heavier than before, single-line and multi-line.
- [ ] Storybook: `<Heading level="2" fontSize="lg">` (or similar override) renders at the token-derived line-height (24px), NOT 30px. Weight stays 700.
- [ ] Design sign-off: @Pei Chien confirms tightening code to match Figma is the correct direction.

## Follow-ups (not in this PR)

- Audit `<Heading level="1|3|4">` against their respective Figma styles — same drift may exist elsewhere in the scale.
- Consider exposing `lineHeight` as a sprinkle prop (or Heading-specific prop) so the current `isDefaultLevel2` guard can be replaced with a general escape hatch. Discussion warranted before landing.

### DIFF
--- a/.changeset/heading-level-2-line-height.md
+++ b/.changeset/heading-level-2-line-height.md
@@ -1,0 +1,8 @@
+---
+"@optiaxiom/react": minor
+"@optiaxiom/web-components": minor
+"@optiaxiom/proteus": minor
+"@optiaxiom/mcp": minor
+---
+
+Aligned `<Heading level="2">` with the Figma `Header/Heading 2` style: tightened `line-height` from `40px` to `30px` and bumped `font-weight` from `700` to `800`. Also added `"800"` to the accepted `fontWeight` sprinkle values (previously the scale jumped `"700"` → `"900"`), which is now reflected in the Proteus JSON schemas and the MCP metadata. Other heading levels and other consumers of `fontSize="3xl"` are unchanged.
--- a/.changeset/heading-level-2-line-height.md
+++ b/.changeset/heading-level-2-line-height.md
@@ -6,3 +6,5 @@
 ---
 
 Aligned `<Heading level="2">` with the Figma `Header/Heading 2` style: tightened `line-height` from `40px` to `30px` and bumped `font-weight` from `700` to `800`. Also added `"800"` to the accepted `fontWeight` sprinkle values (previously the scale jumped `"700"` → `"900"`), which is now reflected in the Proteus JSON schemas and the MCP metadata. Other heading levels and other consumers of `fontSize="3xl"` are unchanged.
+
+**Visual change:** apps using bare `<Heading level="2">` will render tighter and heavier — review level-2 headings for re-flow, clipped descenders, and vertical-rhythm drift before upgrading. To fully opt out on a case-by-case basis, pass an explicit `fontSize`, which restores both the `700` weight and the token's `40px` line-height. Passing only an explicit `fontWeight` overrides the weight but leaves the tighter `30px` line-height in place.
--- a/packages/mcp/src/data.json
+++ b/packages/mcp/src/data.json
@@ -138593,6 +138593,7 @@
       "bg.information": "#036988",
       "bg.information.light": "#91DBDA",
       "bg.information.subtle": "#E2F5F5",
+      "bg.message.neutral": "#F5F6F7",
       "bg.overlay": "#393E3252",
       "bg.page": "#F9FCF8",
       "bg.pill.default": "#D8E4CB",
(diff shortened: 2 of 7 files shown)


radnan:
already addressed in #1762

PR #1773, opened 2026-09-18
feat(proteus): disable actions while an interaction is pending

radnan:
Closes [OPAL-6356](https://optimizely-ext.atlassian.net/browse/OPAL-6356).

## Problem

`ProteusAction` kept its loading state locally, so every action button in a document was independently clickable. In a confirm/cancel document the user can click **Cancel** while **Delete permanently** is still in flight — firing a second interaction against a record that is already being deleted (and vice versa).

## Change

The pending action's id now lives on the document context (`pendingAction` / `setPendingAction` in `ProteusDocumentShell`):

- The action that fired keeps its spinner (`loading={pendingAction === id}`).
- Every other action in the document is disabled until the event settles.
- Dispatch is wrapped in `try/finally`, so a rejected interaction releases the document instead of leaving the buttons stuck loading forever (the previous code would never clear local `loading` on throw).

Scope is deliberately limited to `Action` buttons — other `onEvent` callers (Question, PillMenu, CardLink) are untouched, so background / `mode: "read"` fetches don't freeze the footer. One document = one busy; cross-card serialization stays a host concern.

No schema change: this is renderer behavior, nothing new is exposed to documents.

### Note on the ticket's shape

OPAL-6356 prescribed a `busy: boolean` on the context alongside the existing local `loading`. This stores the pending action's id instead, so `loading` is derived from `pendingAction === id` rather than being a second piece of state that can disagree with the first. The `busy && !loading` rule the ticket describes is preserved exactly.

## Story & test

`WithAsyncActions` is cut down to the essentials — two actions and an `onInteraction` that resolves after 1s — plus a `play` test covering both directions:

| state | Cancel | Delete permanently |
| --- | --- | --- |
| Delete pending | **disabled** | enabled, spinner |
| after settle | enabled | enabled |
| Cancel pending | enabled, spinner | **disabled** |
| after settle | enabled | enabled |

Verified the test actually guards the fix by reverting the `disabled` expression and confirming it fails on `expect(element).toBeDisabled()`.

`pnpm lint`, `pnpm tsgo -b`, and the storybook vitest run (31 passed) are clean.


.changeset/neat-moons-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@optiaxiom/proteus": patch
```
tanvirstreame: nit: bump to `minor`? This adds two required props (`pendingAction`, `setPendingAction`) to `ProteusDocumentProvider`, which is a consumer-facing symbol per the proteus README. No external consumers today, so `patch` is safe — just semver-honest.

packages/proteus/src/proteus-action/ProteusAction.tsx
```diff
@@ -22,31 +22,34 @@ export function ProteusAction({
...
   ) as ProteusEventHandler;
 
-  const [loading, setLoading] = useState(false);
+  const id = useId();
+  const loading = pendingAction === id;
 
   return (
     <Button
-      disabled={type === "submit" && !valid}
+      disabled={(type === "submit" && !valid) || (!!pendingAction && !loading)}
```
tanvirstreame: nit: extract for readability — `(!!pendingAction && !loading)` is equivalent to "another action is pending":

```tsx
const otherActionPending = !!pendingAction && !loading;
disabled={(type === "submit" && !valid) || otherActionPending}
```

packages/proteus/src/proteus-action/ProteusAction.tsx
```diff
@@ -22,31 +22,34 @@ export function ProteusAction({
...
       justifyContent="center"
       loading={loading}
       onClick={async () => {
-        if (!onClick || loading) {
+        if (!onClick || pendingAction) {
```
tanvirstreame, line 47: nit: one-line comment here so a future reader doesn't "simplify" back to `if (loading)`. The check on `pendingAction` (not `loading`) means a keyboard-activate on the *currently loading* button also early-returns — correct, but non-obvious.

tanvirstreame (commented):
LGTM overall — correct fix, well-scoped, real test coverage. A handful of inline notes, none blocking.

packages/proteus/src/proteus-action/ProteusAction.tsx
radnan: Done — extracted as `otherActionPending` in 80d9b27.

packages/proteus/src/proteus-action/ProteusAction.tsx
radnan: Done in 80d9b27 — comment now spells out that the firing action stays enabled to keep its spinner, so re-activating it has to early-return here too.

packages/proteus/src/proteus-action/ProteusAction.tsx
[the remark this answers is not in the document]
radnan: Not changing this — the override is intentional. `{...props}` is spread last throughout this repo precisely so a document author can take control of anything the component computes, and `disabled` is a documented `Action` prop. A card that declares `disabled` is opting into driving that state itself, the same way it already could over the `type === "submit" && !valid` guard that predates this PR. Forcing the internal value to win would remove a deliberate escape hatch.

tanvirstreame (approved):

